PR #12, opened 2015-09-14
rename injectTransaction to sendRawTransaction

CJentzsch:
in connection with:
https://github.com/ethereum/webthree/pull/37
https://github.com/ethereum/alethzero/pull/59

### DIFF
--- a/libethereum/ClientBase.h
+++ b/libethereum/ClientBase.h
@@ -140,7 +140,7 @@ public:
 	virtual BlockInfo pendingInfo() const override;
 	virtual BlockDetails pendingDetails() const override;
 
-	virtual ImportResult injectTransaction(bytes const& _rlp, IfDropped _id = IfDropped::Ignore) override { prepareForTransaction(); return m_tq.import(_rlp, _id); }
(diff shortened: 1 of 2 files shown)


gavofyork:
naming was well-chosen - it's consistent with injectBlock.

CJentzsch:
There is nothing wrong with the naming. The problem is that there is a function called `sendRawTransaction` defined in web3.js, which makes a JSON-RPC call to `eth_sendRawTransaction`. When opening the js console, one can not use the method (Method not found error) although the output of `web3` would suggest so https://github.com/ethereum/webthree/issues/36.
Go has this function implemented as JSON-RPC call, and there are already libs demanding it (and therefore have to use a go node instead of a cpp node). Cpp has the same functionality in `injectTransaction` . Its just about the name.
The following solutions exist:
1. Go and web3.js change the name
2. Cpp changes the name
3. Cpp leaves `injectTransaction` and implements `sendRawTransaction` additionally, although they do the same thing.

I did choose 2 (ethereum/webthree#37) . But let me know what you prefer.

Usually we just follow the definition in https://github.com/ethereum/wiki/wiki/JSON-RPC, but this function is not defined yet, but both clients have implemented it with a different name.

PR #14, opened 2015-09-14
optional olympic difficulty algorithm

CJentzsch:
Without it, one can not sync with the olympic network. And for private networks it is useful.


libethcore/BlockInfo.cpp
```diff
@@ -208,6 +208,8 @@ u256 BlockInfo::calculateDifficulty(BlockInfo const& _parent) const
 	if (!m_number)
 		throw GenesisBlockCannotBeCalculated();
 	u256 o = max<u256>(c_minimumDifficulty, m_timestamp >= _parent.m_timestamp + c_durationLimit ? _parent.m_difficulty - (_parent.m_difficulty / c_difficultyBoundDivisor) : (_parent.m_difficulty + (_parent.m_difficulty / c_difficultyBoundDivisor)));
+	if (c_network != Network::Frontier)
```
gavofyork: `c_network == Network::Olympic`?

CJentzsch: Thats what I had before, but the same rule applies for Turbo (we didn't have murden at the time for the PR. So either `c_network == Network::Olympic || c_network == Network::Turbo` or `c_network != Network::Frontier`
But I will change that, since the new diff algo applies to morden as well.

PR #22, opened 2015-09-17
KeyManager changeName method

debris:


libethcore/KeyManager.cpp
```diff
@@ -342,6 +342,19 @@ string const& KeyManager::accountName(Address const& _address) const
...
+{
+	try
+	{
+		m_keyInfo.at(_address).accountName = _name;
+	}
+	catch (...)
+	{
+		return;
+	}
+	write(m_keysFile);
```
CJentzsch: Write can fail. If so, the name should not be changed.

debris: I assumed that it will not fail. Like it's done with all the other methods in this file.

CJentzsch: Sorry, but I am not convinced that this function can not throw. 
In `KeyManger::write` the function `writeFile` is called, where is this function defined?

libethcore/KeyManager.cpp
```diff
@@ -342,6 +342,19 @@ string const& KeyManager::accountName(Address const& _address) const
 	}
 }
 
+void KeyManager::changeName(Address const& _address, std::string const& _name)
+{
+	try
+	{
+		m_keyInfo.at(_address).accountName = _name;
+	}
+	catch (...)
```
chfast: Why do you use exceptions as a replacement for control flow structures? Use `find` and `if`.

libethcore/KeyManager.cpp
```diff
@@ -342,6 +342,19 @@ string const& KeyManager::accountName(Address const& _address) const
...
+	{
+		return;
```
chfast: This function silently does nothing in case of invalid input. Are you sure it is ok in this case? In general it is _very not ok_.

debris: you are right, I shouldn't copynpaste. Will fix it.

debris: I fixed it in latest pr.
Other methods in this class also do not throw on invalid input.

libethcore/KeyManager.cpp
debris: CommonIO.
I'm saying that, this method behaves exactly like any other method in this class. If write fails, it will fail in all places where methods using `write` are used. I don't want to make changing/handling this behaviour part of this pr.

libethcore/KeyManager.cpp
```diff
@@ -342,6 +342,15 @@ string const& KeyManager::accountName(Address const& _address) const
 	}
 }
 
+void KeyManager::changeName(Address const& _address, std::string const& _name)
+{
+	if (m_keyInfo.count(_address))
+	{
+		m_keyInfo.at(_address).accountName = _name;
```
chfast: You don't have to search twice. Use `find`.

debris: `m_keyInfo` is a map. It won't. Will always return 0 or 1.

chfast: `.count` is the first map search, `.at` is the second one.

chfast:
```
  auto it = m_keyInfo.find(_address);
  if (it != m_keyInfo.end())
    it->second = _name;
```

chfast:
Perfect. Thanks.

arkpar:
looksgood

PR #28, opened 2015-09-21
Remove libWeb3Tests

winsvega:
Moved to libweb3core repo https://github.com/ethereum/libweb3core/pull/10


winsvega:
https://github.com/ethereum/libweb3core/pull/10 (move with history)

test/libweb3core/crypto.cpp
```diff
@@ -0,0 +1,150 @@
+/*
+	This file is part of cpp-ethereum.
```
chfast, line 2: Where does this file come from?

test/libweb3core/crypto.cpp
```diff
@@ -0,0 +1,150 @@
...
+		cnote << toHex(rlp);
+		cnote << t.sha3(eth::WithoutSignature);
+		rlp = t.rlp(eth::WithSignature);
+		cnote << RLP(rlp);
+		cnote << toHex(rlp);
+		cnote << t.sha3(eth::WithSignature);
+		assert(t.sender() == p.address());
+	}
+
+#if 0
```
chfast: Can we remove this disabled code?

test/libweb3core/crypto.cpp
winsvega, line 2: this are the tests for libweb3core that have dependencies on libethereum. thats why it is in libethereum.

PR #29, opened 2015-09-22
Create version string.

chriseth:


libevmasm/Version.cpp
```diff
@@ -0,0 +1,38 @@
...
+#include <libevmasm/Version.h>
+#include <string>
+#include <ethereum/BuildInfo.h>
+#include <libdevcore/Common.h>
+
+using namespace dev;
+using namespace std;
+
+char const* dev::eth::VersionNumberLibEvmAsm = ETH_PROJECT_VERSION;
+extern string const dev::eth::VersionStringLibEvmAsm =
```
chfast, line 32: `extern` is incorrect here.

chfast, line 32: If you use `using namespace dev` everywhere do not repeat it here again.

libevmasm/Version.cpp
```diff
@@ -0,0 +1,38 @@
+/*
+	This file is part of cpp-ethereum.
```
chfast, line 2: cpp-ethereum is dead.

PR #37, opened 2015-10-05
Moved key generation code to KeyManager

arkpar:
This was moved from AlethZero


arkpar:
https://github.com/ethereum/alethzero/pull/79

PR #40, opened 2015-10-06
Evmjit submodule

chfast:


bobsummerwill:
Oh yes, please!
@chriseth @LefterisJP @subtly @chfast.

chfast:
I would like to have it too very much @bobsummerwill. Unfortunately, it seems the owners are not interested any more.

bobsummerwill:
More like de-resourced than uninterested, I think, @chfast :-(

bobsummerwill:
Hey @chriseth!

Looking at the dependency graph - https://camo.githubusercontent.com/b8decc1517241cfe67e05d4dce4ca7fd1b27c03d/687474703a2f2f646f75626c657468696e6b636f2e6769746875622e696f2f77656274687265652d756d6272656c6c612d63726f73732f696d616765732f646570656e64656e63795f67726170682e737667

I think that separating evmasm/evmcore into their own git repo would be very beneficial for solidity too, eh?    I don't know whether we would want it as a sub-module under libethereum, or just as a sub-module under webthree-umbrella, but doing that would uncouple libsolidity from the rest of the libethereum repo.

What kind of strength of dependency on libethereum do the solidity tests have?

CC @CJentzsch.

bobsummerwill:
See also https://github.com/ethereum/webthree-umbrella/issues/65, for an earlier evmasm/evmasmcore removal suggestion.

This issue is specially talking about evmjit, which I think also makes sense, but would have to happen AFTERWARDS.

chfast:
I'm not sure anyone is managing this repo any more.

As I remember the problem were buildbots not handling this layout correctly.

chriseth:
I don't see the dependency problem as an argument for moving to a submodule. This is only a matter of cmake configuration. In the long term, solidity end to end tests should be done in a way that is less tightly linked to the VM anyway.

chfast:
The only problem I see here is this code is almost identical with https://github.com/ethereum/evmjit/tree/develop. Managing both repos is a waste of time.

@bobsummerwill why do you actually need this?

bobsummerwill:
I didn't spotted https://github.com/ethereum/evmjit/tree/develop :-(
What is that repo used for?
I don't _need_ this, but would like to see Solidity decoupled from the rest of libethereum.
I think that EVMJIT also makes sense to pull out, because the LLVM dependency is so heavy.

chfast:
Historically, EVMJIT was separated from beginning, but later pulled as a git subtree merge (means copy and paste). etherem/evmjit repo is the same code as here under libethereum/evmjit.

You can still build without it setting `-DEVMJIT=0` to cmake.

bobsummerwill:
I'm obviously missing some history here, which maybe you guys can fill me in on.

What was the reason for the cut-and-paste?     https://github.com/ethereum/evmjit is building a library?

So we've got two copies of the code now?   We shouldn't have two copies.

chfast:
I'm happy to do the split work again to have evmjit as a submodule. That would be much better for me. But the rest of developers must agree on that.

bobsummerwill:
Hey @chriseth!   Do you know what the rationale was for cut-and-pasting in the first place?

chriseth:
Moving evmjit to its own repo makes much more sense than moving anything else to its own repo. Especially considering that we plan to make an evmjit library that has a clear C interface.
Although, this needs some changes to the build system and we currently need to focus on different things, but we might start doing that soon in concert with the geth team. I will create a story in our backlog for that and close this pull request for now because it is quite outdated (and also to clear the list of open pull requests).

chfast:
Understandable.

In the future, consider adding evmjit submodule on top level, next to libethereum.

chriseth:
Yes, submodules should not be hidden in subdirectories or other submodules.

bobsummerwill:
Yes - we shouldn't nest them.    A top-level module would make sense to me.

Do you guys think it might make sense to group evmcore, evmasm and evmjit into a single repository?

I know that we have the dependencies from Solidity onto evmasm.    What other clients are there for the EVM code, other than Solidity and libethereum?    How is evmjit used standalone?

chriseth:
I think it would be better to move evmasm (and lllc) into solidity. evmjit already is pretty well isolated (great job, @chfast !). It should be possible to use it as a shared library so that you call a function where you supply bytecode and callbacks for environmental info (storage, calling other contracts, ...) and it returns the result of the computation.

It could be possible to make evmcore header-only and thus completely remove the link-dependency from solc onto libethereum (the tests still depend on it for now).

chfast:
EVM JIT depends only on LLVM. That was the goal from the beginning to make integration with other languages easier.

I would like to provide also a small library implementing plugin-like loader for evmjit.so. In this case EVM JIT could be built independently and distributed as a library in deb/homebrew/etc.

chriseth:
Actually do you think it might be possible to create a stand-alone evmjit executable and communicate with it via streams? The thing is that the isolation of interpreted code is much higher and the possibility for evm code to break out of isolation also depends on possible bugs in llvm. If that happens, it might not only create a hard fork but might also compromise private keys.

chfast:
I was thinking about it. I don't know much about IPC performance. It might kill any gains. The first step would be allowing JIT to execute full call stack instead of executing calls in separation.

PR #42, opened 2015-10-07
Wallet compatibility fixes

arkpar:
This PR sets gas for eth_call to maximum regardless of calling account balance


libethereum/ClientBase.cpp
```diff
@@ -70,6 +70,8 @@ ExecutionResult ClientBase::call(Address const& _from, u256 _value, Address _des
 	{
 		Block temp = asOf(_blockNumber);
 		u256 n = temp.transactionsFrom(_from);
+		if (_ff == FudgeFactor::Lenient)
```
gavofyork: this will almost certainly break existing code. `_gas` should be respected, even (especially) when lenient.

PR #45, opened 2015-10-12
State tree pruning

CJentzsch:
To be merged together with: https://github.com/ethereum/libweb3core/pull/21 (build fails because of this)

Until the new patricia merkle tree is implemented the expensive tree check in Block::cleanup has to be made.
Not ready for release, but as "experimental" feature.


libethereum/Block.cpp
```diff
@@ -805,7 +816,44 @@ void Block::cleanup(bool _fullCommit)
 			throw;
 		}
 
-		m_state.db().commit();	// TODO: State API for this?
+		m_state.db().commit(m_currentBlock.number());	// TODO: State API for this?
+
+		try
+		{
+			if (m_currentBlock.number() % Defaults::statePruning() == 0)
+				mutableState().getState().leftOvers();
```
CJentzsch, line 824: this as well as line 844 is a workaround which should be removed later. It just scans the whole tree and tries to lookup every node. In the case a node gets lookedup and has a reference count of zero, the reference count gets increased by one and it is removed from the deathRow (the list of nodes which will be deleted in X blocks)

PR #46, opened 2015-10-12
Better long multiplication algorithm.

chfast:


evmjit/libevmjit/Arith256.cpp
```diff
@@ -23,68 +23,114 @@ Arith256::Arith256(IRBuilder& _builder) :
 	CompilerHelper(_builder)
 {}
 
-void Arith256::debug(llvm::Value* _value, char _c)
+void Arith256::debug(llvm::Value *_value, char _c, llvm::Module &_module, IRBuilder &_builder)
```
chriseth: `*` and `&` should be "left-aligned" (also in other places).

chfast: Hmm... where that came from?

PR #51, opened 2015-10-14
Better test running output to cmd

winsvega:


test/external-dependencies/boost.cpp
```diff
@@ -22,12 +22,14 @@
...
 #include <boost/test/unit_test.hpp>
 #include <libdevcore/Common.h>
+#include <test/TestHelper.h>
 
 BOOST_AUTO_TEST_SUITE(ExtDepBoost)
 
 // test that reproduces issue https://github.com/ethereum/cpp-ethereum/issues/1977
 BOOST_AUTO_TEST_CASE(u256_overflow_test)
 {
+	dev::test::TestOutputHelper::initTest();
```
CJentzsch, line 32: I see that this is called in every test. Isn't there some functionality/option in the boost test library that you can use?
I guess they have test reporting tool integrated and you don't need to build your own, rather use theirs.

winsvega, line 32: This line basically jsut replaces  cnote<<testCaseName; 

is some BOOST cases. such as BlockChain, Vm, State,  Transaction 
we use a different method to output the test case name.
thus I have TestOutputHelper::initTest(json::mObject);
and TestOutputHelper::initTest();

I think boost does not allow to add a user defined function to be called before each boost test case function

CJentzsch, line 32: I still can't imagine that boost test is supposed to work that way, that you need to `cnote << testname` ot use your helper. I don think you need to add a user defined function, just check the commandline flags or the docs (http://www.boost.org/doc/libs/1_43_0/libs/test/doc/html/utf/user-guide/test-output.html)

winsvega, line 32: there is actually boost cmd option. --log_level=message 
but it does not output every test case name 

other boost options would output too many information. 
you wanted that with --verbosity 1 test to output only test case name and errors (if any)

PR #54, opened 2015-10-16
BlockQueue thread lock issue fix

winsvega:
should fix that bug 
https://github.com/ethereum/webthree-umbrella/issues/43

### DIFF
--- a/libethereum/BlockQueue.cpp
+++ b/libethereum/BlockQueue.cpp
@@ -66,11 +66,10 @@ BlockQueue::~BlockQueue()
 
 void BlockQueue::stop()
 {
-	{
-		std::lock_guard<std::mutex> l(m_verification);
+	DEV_GUARDED(m_verification)
 		m_deleting = true;
-		m_moreToVerify.notify_all();
-	}
+
+	m_moreToVerify.notify_all();
 	for (auto& i: m_verifiers)
 		i.join();
 	m_verifiers.clear();
--- a/libethereum/BlockQueue.cpp
+++ b/libethereum/BlockQueue.cpp
@@ -66,8 +66,11 @@ BlockQueue::~BlockQueue()
 
 void BlockQueue::stop()
 {
-	m_deleting = true;
-	m_moreToVerify.notify_all();
+	{
+		std::lock_guard<std::mutex> l(m_verification);
+		m_deleting = true;
+		m_moreToVerify.notify_all();
+	}
 	for (auto& i: m_verifiers)
 		i.join();
 	m_verifiers.clear();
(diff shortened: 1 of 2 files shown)


gavofyork:
could you explain how this fixes the bug?

winsvega:
> The condition_variable class is a synchronization primitive that can be used to block a thread, or multiple threads at the same time, until another thread both modifies a shared variable (the condition), and notifies the condition_variable.
> 
> The thread that intends to modify the variable has to
> 
>   acquire a std::mutex (typically via std::lock_guard)
>     perform the modification while the lock is held
>     execute notify_one or notify_all on the std::condition_variable (the lock does not need to be held for notification) 
> 
> Even if the shared variable is atomic, it must be modified **under the mutex** in order to correctly publish the modification to the waiting thread. 

I noticed that m_deleting have been modified without acquiring the mutex. It seems it caused the threads not joining at destructor ~once per 5000 destructor operations.  I tested this patch after doing the modifications under the mutex and never get this bug again even on 10000 destructor operations.

gavofyork:
ok so the patch currently does three things:
- change type of `m_deleting` to atomic;
- protect writing to `m_deleting` with mutex;
- protect notification with same mutex.

in fact, the fix is only the second point, to protect the writing to `m_deleting` with mutex, the change to atomic and inclusion of the notification `m_moreToVerify.notify_all();` within the guard's scope is unnecessary. could you check the bug is still fixed if these are removed?

libethereum/BlockQueue.cpp
```diff
@@ -66,8 +66,11 @@ BlockQueue::~BlockQueue()
 
 void BlockQueue::stop()
 {
-	m_deleting = true;
-	m_moreToVerify.notify_all();
+	{
+		std::lock_guard<std::mutex> l(m_verification);
+		m_deleting = true;
+		m_moreToVerify.notify_all();
```
gavofyork: move outside scope

libethereum/BlockQueue.cpp
```diff
@@ -66,8 +66,11 @@ BlockQueue::~BlockQueue()
 
 void BlockQueue::stop()
 {
-	m_deleting = true;
-	m_moreToVerify.notify_all();
+	{
+		std::lock_guard<std::mutex> l(m_verification);
```
gavofyork: use:

``` c++
DEV_GUARDED(m_verification)
    m_deleting = true;
```

instead.

libethereum/BlockQueue.h
```diff
@@ -162,7 +162,7 @@ class BlockQueue: HasInvariants
 	std::deque<UnverifiedBlock> m_unverified;							///< List of <block hash, parent hash, block data> in correct order, ready for verification.
 
 	std::vector<std::thread> m_verifiers;								///< Threads who only verify.
-	bool m_deleting = false;											///< Exit condition for verifiers.
+	std::atomic<bool> m_deleting = { false };							///< Exit condition for verifiers.
```
gavofyork: unneeded.

winsvega:
> could you check the bug is still fixed if these are removed?

no. I have no way to reproduce this bug 100%.
all I could do is to gather some statistic that this bug wont happen on a long run.

but I still guess it is better be guarded by
DEV_GUARDED(m_verification)

PR #55, opened 2015-10-19
Implement GetBlockHeaders message

chfast:


libethereum/EthereumPeer.cpp
```diff
@@ -246,22 +246,80 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
...
-		for (unsigned i = 0; i < c && p; ++i, p = host()->chain().details(p).parent)
-			s << p;
+		prep(s, BlockHeadersPacket, numHeadersToSend);
+		for (unsigned i = 0; i != numHeadersToSend; ++i)
+		{
+			if (!hash)
+				break;
+
+			s << host()->chain().headerData(hash);
+			hash = host()->chain().details(hash).parent;
```
arkpar: `skip` and `reverse` should be used here instead of simple child to parent traversal

libethereum/EthereumPeer.cpp
```diff
@@ -246,22 +246,80 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
...
-		h256 p = host()->chain().details(later).parent;
-		for (unsigned i = 0; i < c && p; ++i, p = host()->chain().details(p).parent)
-			s << p;
+		prep(s, BlockHeadersPacket, numHeadersToSend);
+		for (unsigned i = 0; i != numHeadersToSend; ++i)
+		{
+			if (!hash)
+				break;
+
+			s << host()->chain().headerData(hash);
```
arkpar: I think you need to use `s.appendRaw` here, otherwise RLP will be encoded twice

libethereum/EthereumPeer.cpp
chfast: `skip` and `reverse` determine initial value of `hash`. We could iterate using block numbers but that not ideal as we cannot lock the block chain for iteration. Hashes are safer, but not ideal too.

arkpar: `skip` and `reverse` should not affect initial value, but how following block numbers are calculated. You  can still do enumeration by hashes, although it will be inefficient for large `skip`

libethereum/EthereumPeer.cpp
```diff
@@ -37,8 +37,10 @@ using namespace dev;
 using namespace dev::eth;
 using namespace p2p;
 
-const unsigned c_maxIncomingNewHashes = 1024;
-const unsigned c_maxHashesToSend = 65536;
+static const unsigned c_maxIncomingNewHashes = 1024;
+static const unsigned c_maxHeadersToSend = 65536;
+static const unsigned c_maxBlockSkip = 99999;
+static_assert(c_maxHeadersToSend * (c_maxBlockSkip + 1) > c_maxHeadersToSend, "Multiplication of config constants overflows");
```
arkpar: What does it check for? Assert expression does overflow currently

libethereum/EthereumPeer.cpp
```diff
@@ -37,8 +37,10 @@ using namespace dev;
 using namespace dev::eth;
 using namespace p2p;
 
-const unsigned c_maxIncomingNewHashes = 1024;
-const unsigned c_maxHashesToSend = 65536;
+static const unsigned c_maxIncomingNewHashes = 1024;
+static const unsigned c_maxHeadersToSend = 65536;
+static const unsigned c_maxBlockSkip = 99999;
```
arkpar: Don't see much sense in limiting it

libethereum/EthereumPeer.cpp
```diff
@@ -37,8 +37,10 @@ using namespace dev;
 using namespace dev::eth;
 using namespace p2p;
 
-const unsigned c_maxIncomingNewHashes = 1024;
-const unsigned c_maxHashesToSend = 65536;
+static const unsigned c_maxIncomingNewHashes = 1024;
+static const unsigned c_maxHeadersToSend = 65536;
```
arkpar: Should be much lower, I suggest 1024. Header size is considerably larger than hash. Also, there is 16mb packet size limit.

libethereum/EthereumPeer.cpp
chfast: `skip` requires linear traversal by block headers currently. Big values would require to fetch whole db. Remove the limit?

libethereum/EthereumPeer.cpp
chfast: What is the max RLP encoded header size?

libethereum/EthereumPeer.cpp
```diff
@@ -246,22 +248,77 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
...
+			}
+		}
+		else // block id is a number
+		{
+			auto n = blockId.toInt<bigint>();
+			clog(NetMessageSummary) << "GetBlockHeaders (block (number): " << n
+			<< ", maxHeaders: " << maxHeaders
+			<< ", skip: " << skip << ", reverse: " << reverse << ")";
+
+			if (reverse)
```
arkpar: `if (!reverse)` ?

chfast: We cannot traverse forward (`!reverse`), so I'm looking for top requested block.

arkpar: Exactly. But the condition is inverted, isn't it?

libethereum/EthereumPeer.cpp
arkpar: ~540 bytes

libethereum/EthereumPeer.cpp
```diff
@@ -246,22 +246,134 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
...
+		auto numHeadersToSend = maxHeaders <= c_maxHeadersToSend ? static_cast<unsigned>(maxHeaders) : c_maxHeadersToSend;
+
+		if (skip > std::numeric_limits<unsigned>::max() - 1)
+		{
+			clog(NetAllDetail) << "Requested block skip is too big: " << skip;
+			break;
+		}
+
+		auto step = static_cast<unsigned>(skip) + 1;
+		assert(step > 0 && "step must not be 0");
```
arkpar, line 269: assert will never be triggered

chfast, line 269: It should not. It's an assert.

arkpar, line 269: What i mean is that `step` can't be 0 here since it is already checked by if clause above

chfast, line 269: I agree. The assert is trivial. I will think about removing it.

PR #64, opened 2015-10-21
Block.cpp Coverage

winsvega:
report https://github.com/ethereum/webthree-umbrella/issues/74

### DIFF
--- a/test/BlockChainHelper.cpp
+++ b/test/BlockChainHelper.cpp
@@ -192,7 +192,10 @@ void TestBlock::mine(TestBlockChain const& _bc)
 
 		premineUpdate(blockInfo);
 
+		size_t transactionsOnImport = m_transactionQueue.topTransactions(100).size();
 		block.sync(blockchain, m_transactionQueue, gp); //!!! Invalid transactions are dropped here
+		if (transactionsOnImport >  m_transactionQueue.topTransactions(100).size())
+			cnote << "Dropped invalid Transactions when mining!";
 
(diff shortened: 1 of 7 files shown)


test/BlockChainHelper.cpp
```diff
@@ -462,13 +462,13 @@ vector<TestBlock> TestBlockChain::syncUncles(vector<TestBlock> const& uncles)
 	return validUncles;
 }
 
-TestTransaction TestTransaction::getDefaultTransaction()
+TestTransaction TestTransaction::getDefaultTransaction(string const& _nonce, string const& _gasLimit, string const& _data)
```
CJentzsch: Although you need a string in the function, those values are clearly numbers. Use u256, u64, bytes ...  and convert to string in the function.

test/libethereum/block.cpp
```diff
@@ -0,0 +1,232 @@
...
+
+	try
+	{
+		Block block32 = blockchain.genesisBlock(genesisDB);
+		block32.populateFromChain(blockchain, h256("0x0000000000000000000000000000000000000000000000000000000000000001"));
+		BOOST_ERROR("Expected BlockNotFound exception!");
+	}
+	catch (std::exception const& _e)
+	{
+		BOOST_REQUIRE(string(_e.what()).find("BlockNotFound") != string::npos);
```
CJentzsch: use `BOOST_CHECK_EXCEPTION` (http://www.boost.org/doc/libs/1_35_0/libs/test/doc/components/test_tools/reference/BOOST_CHECK_EXCEPTION.html) instead

PR #67, opened 2015-10-22
Response to GetBlockBodies/BlockBodies

chfast:


libethereum/EthereumPeer.cpp
```diff
@@ -397,44 +397,51 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
...
 			if (host()->chain().isKnown(h))
 			{
-				rlp += host()->chain().block(_r[i].toHash<h256>());
+				auto blockBytes = host()->chain().block(h);
+				RLP block{blockBytes};
+				auto transactions = block[1].data();
+				auto uncles = block[2].data();
+				auto inserter = std::back_inserter(rlp);
+				std::copy(transactions.begin(), transactions.end(), inserter);
+				std::copy(uncles.begin(), uncles.end(), inserter);
```
arkpar: RLP should have the structure described in the spec: a list of items, where each element is a list of two items: transaction and uncles. There is no need to iterate over transactions and uncles, just copy raw data

chfast: I'm actually coping bytes from RLP of list of transactions, not transactions as RLP items. But I can see that a list containing 2 items is missing there.

What do you mean by "copy raw data" exactly?

PR #68, opened 2015-10-22
Add an option to disable eth_submitHashrate RPC call on node to ethminer

sammy007:
I added an option to ethminer to disable eth_submitHashrate RPC call to node, because it generate a lot of useless traffic, pools don't use it as far as I know and it always show wrong numbers anyway.

<code>--disable-submit-hashrate  When mining, don't submit hashrate to node.</code>

### DIFF
--- a/ethminer/MinerAux.h
+++ b/ethminer/MinerAux.h
@@ -287,6 +287,8 @@ public:
 				BOOST_THROW_EXCEPTION(BadArgument());
 			}
 		}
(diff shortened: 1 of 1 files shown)


sammy007:
For those who mining solo and want to track rigs availability along with real hashrate I recommend my https://github.com/sammy007/ether-proxy. Please add to wiki if it's allowed. Thanks.

PR #69, opened 2015-10-23
PV63: GetNodeData

chfast:


libethereum/EthereumPeer.cpp
```diff
@@ -474,6 +474,40 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
...
+		unsigned count = static_cast<unsigned>(_r.itemCount());
+		if (!count)
+		{
+			clog(NetImpolite) << "Zero-entry GetNodeData: Not replying.";
+			addRating(-10);
+			break;
+		}
+		clog(NetMessageSummary) << "GetNodeData (" << dec << count << "entries)";
+
+		// return the requested blocks.
```
arkpar: forgot to update comment

libethereum/EthereumPeer.cpp
```diff
@@ -474,6 +474,40 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
...
+		bytes rlp;
+		unsigned n = 0;
+		auto numItemsToSend = std::min(count, c_maxBlocks);
```
arkpar: better use a new constant

chfast:
@arkpar comments addressed.

libethereum/EthereumPeer.cpp
```diff
@@ -474,6 +474,40 @@ bool EthereumPeer::interpret(unsigned _id, RLP const& _r)
...
+		bytes rlp;
+		unsigned n = 0;
+		auto numItemsToSend = std::min(count, c_maxNodes);
+		for (unsigned i = 0; i < numItemsToSend && rlp.size() < c_maxPayload; ++i)
+		{
+			auto h = _r[i].toHash<h256>();
+			auto node = host()->db().lookup(h);
+			if (!node.empty())
+			{
+				auto nodeData = RLPStream{}.append(node).out();
```
arkpar: I think node data is already RLP encoded, no need to encode it twice

chfast:
I've removed node data double encoding.
Added reponse to GetReceipts too.

PR #71, opened 2015-10-24
Allow building with the new Boost 1.59 version

LefterisJP:
Boost has changed the internal implementation of boost test, with the new version so our borrowing of boost internals here needed adjusting.

@winsvega Can you take a look that it does not break anything test related?


LefterisJP:
Should fix https://github.com/ethereum/webthree-umbrella/issues/78

gavofyork:
ubuntu not passing?

LefterisJP:
@gavofyork Nope it's just Jenkins master acting up due to insuffficient resources.

PR #76, opened 2015-10-28
CanonBlockChain coverage tests

winsvega:
Report: ethereum/webthree-umbrella#74


test/libethereum/canonblockchain.cpp
```diff
@@ -0,0 +1,119 @@
...
+#include <libethereum/CanonBlockChain.h>
+#include <libdevcore/TransientDirectory.h>
+#include <libethereum/Block.h>
+#include <test/TestHelper.h>
+
+using namespace std;
+using namespace dev;
+using namespace dev::eth;
+
+std::string const c_testGenesis = R"(
```
CJentzsch, line 32: Why don't you use: https://github.com/ethereum/libethereum/blob/develop/libethereum/GenesisInfo.cpp#L28 ?

winsvega, line 32: this is for testing blockchain setGenesis()  function.
I have to know what values to expect for defined genesis. 

see report on this coverage:  ethereum/webthree-umbrella#74

PR #79, opened 2015-11-03
Repair eth_syncing

arkpar:


libethereum/ClientBase.cpp
```diff
@@ -140,7 +140,7 @@ std::pair<u256, ExecutionResult> ClientBase::estimateGas(Address const& _from, u
 			State tempState(block.state());
 			tempState.addBalance(_from, (u256)(t.gas() * t.gasPrice() + t.value()));
 			er = tempState.execute(env, t, Permanence::Reverted).first;
-			if (er.excepted == TransactionException::OutOfGas || er.excepted == TransactionException::OutOfGasBase || er.excepted == TransactionException::OutOfGasIntrinsic || er.codeDeposit == CodeDeposit::Failed)
+			if (er.excepted == TransactionException::OutOfGas || er.excepted == TransactionException::OutOfGasBase || er.excepted == TransactionException::OutOfGasIntrinsic || er.codeDeposit == CodeDeposit::Failed || er.excepted == TransactionException::BadJumpDestination)
```
LefterisJP: It would be nice to separate this huge if with each condition on its own line

PR #85, opened 2015-11-06
framing compatibility (interface changed)

gluk256:


gluk256:
Important note: should be merged together with two other PRs, otherwise would not compile.

PR #86, opened 2015-11-06
small test fix and genesis check

winsvega:


test/BlockChainHelper.cpp
```diff
@@ -343,7 +346,7 @@ void TestBlock::recalcBlockHeaderBytes(RecalcBlockHeader _recalculate)
 	}
 
 	RLPStream blHeaderStream;
-	m_blockHeader.streamRLP(blHeaderStream, WithSeal);
+	m_blockHeader.streamRLP(blHeaderStream, WithoutSeal);
```
gavofyork: there should be no need to use `WithoutSeal` - rather the `SealEngine` should be `NoProof` and so the eventual block will not have any seal fields, even when "sealed".

PR #88, opened 2015-11-09
add removed unique_ptr.get() back

LefterisJP:
DONTBUILD

The get() is needed here because llvm::MemoryBuffer::getFile() does not
actually return a simple unique_ptr but instead returns an `ErrorOr`
construct and as such the compiler needs to first get the pointer and
then get the function of the file. Else it does not compile.

Here is the error I was getting

```
[ 13%] Building CXX object evmjit/libevmjit/CMakeFiles/evmjit.dir/Endianness.cpp.o
/home/lefteris/ew2/libethereum/evmjit/libevmjit/Cache.cpp:121:58: error: no member named 'getBuffer' in 'std::unique_ptr<llvm::MemoryBuffer, std::default_delete<llvm::MemoryBuffer> >'
                g_lastObject = llvm::MemoryBuffer::getMemBufferCopy(r->getBuffer());
```

Here is the function signature

```
  static ErrorOr<std::unique_ptr<MemoryBuffer>>
  getFile(const Twine &Filename, int64_t FileSize = -1,
          bool RequiresNullTerminator = true, bool IsVolatileSize =
          false);
```

Plus other fixes in the next commits.

### DIFF
--- a/evmjit/libevmjit/Cache.cpp
+++ b/evmjit/libevmjit/Cache.cpp
@@ -118,7 +118,7 @@ std::unique_ptr<llvm::Module> Cache::getObject(std::string const& id)
 	llvm::sys::path::append(cachePath, id);
 
 	if (auto r = llvm::MemoryBuffer::getFile(cachePath, -1, false))
-		g_lastObject = llvm::MemoryBuffer::getMemBufferCopy(r->getBuffer());
+		g_lastObject = llvm::MemoryBuffer::getMemBufferCopy(r.get()->getBuffer());
 	else if (r.getError() != std::make_error_code(std::errc::no_such_file_or_directory))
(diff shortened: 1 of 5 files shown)


LefterisJP:
@gavofyork One extra problem that I am not 100% sure how to address since git shows this is a new file is here:
https://github.com/ethereum/libethereum/blob/hot_gav/libethashseal/EthashClient.cpp#L61

``` c++
Ethash* EthashClient::sealEngine()
```

seems to always call itself which would lead to an infinite recursion. What's the idea behind it? What was the original purpose?

gavofyork:
should be `return dynamic_cast<Ethash*>(Client::sealEngine());`

LefterisJP:
Ah okay let me add it then

EDIT: Done, added. Thanks.

PR #95, opened 2015-11-20
Fixes and enhancements for sealing/mining

gavofyork:


libethereum/Block.cpp
```diff
@@ -790,10 +812,18 @@ bool Block::sealBlock(bytesConstRef _header)
...
 	// TODO: Leave this in a better state than this limbo, or at least record that it's in limbo.
-	m_transactions.clear();
+	/*m_transactions.clear();
 	m_receipts.clear();
 	m_transactionSet.clear();
-	m_precommit = m_state;
+	m_precommit = m_state = fromPending(0);*/
+	m_state = m_precommit;
+
+	// Commenting out the above probably broke something else:
```
LefterisJP: Is this figured out yet? The code seems to be still commented out

libethereum/TransactionQueue.cpp
```diff
@@ -355,7 +356,7 @@ void TransactionQueue::dropGood(Transaction const& _t)
 	makeCurrent_WITH_LOCK(_t);
 	if (!m_known.count(_t.sha3()))
 		return;
-	m_dropped.insert(_t.sha3());
+//	m_dropped.insert(_t.sha3());
```
LefterisJP: Remove this?

libethereum/TransactionQueue.cpp
```diff
@@ -35,8 +35,8 @@ const size_t c_maxVerificationQueueSize = 8192;
 
 TransactionQueue::TransactionQueue(unsigned _limit, unsigned _futureLimit):
 	m_current(PriorityCompare { *this }),
-	m_limit(_limit),
-	m_futureLimit(_futureLimit)
+	m_limit(_limit * 10),
```
CJentzsch: that is very unpredictable. A constructor should  set those numbers directly as expected.

libethcore/Precompiled.cpp
```diff
@@ -42,6 +41,39 @@ PrecompiledExecutor const& PrecompiledRegistrar::executor(std::string const& _na
 namespace
 {
 
+ETH_REGISTER_PRECOMPILED(ecrecoverOrig)(bytesConstRef _in, bytesRef _out)
```
arkpar: Is this used anywhere? Also I can't spot a difference with the function below

PR #114, opened 2015-12-01
Set sealEngine in tests

CJentzsch:


test/TestHelper.cpp
```diff
@@ -126,7 +126,7 @@ bytes ImportTest::executeTest()
 	eth::State tmpState = m_statePre;
 	try
 	{
-		unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(Network::FrontierTest)).createSealEngine());
+		unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(Options::get().sealEngine)).createSealEngine());
```
winsvega: better use sealEngineNetwork

test/libethereum/TransactionTests.cpp
```diff
@@ -35,7 +35,7 @@ namespace dev {  namespace test {
 
 void doTransactionTests(json_spirit::mValue& _v, bool _fillin)
 {
-	unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(Network::FrontierTest)).createSealEngine());
+	unique_ptr<SealEngineFace> se(ChainParams(genesisInfo(Options::get().sealEngine)).createSealEngine());
```
winsvega: Also change a line in BlockChainHelper.cpp line 444
that would be enough to make all\* blockchain tests to run on defined sealengine

PR #122, opened 2015-12-03
Using set(EXECUTABLE testeth) is tests for libethereum, for consisten…

bobsummerwill:
…cy with other CMake files, and so that automated dependency generation works.


bobsummerwill:
WOOT!    My first commit into Ethereum, @gavofyork :-)

PR #123, opened 2015-12-04
fix test fillers

winsvega:


CJentzsch:
All tests fail, since this depends on the sealengine used. You need to make it sealengine dependent.

PR #128, opened 2015-12-10
State callcodes value transfer test

winsvega:
State callcodes value transfer test


test/libethereum/StateTestsFiller/stCallCodesFiller.json
```diff
@@ -1,4 +1,6 @@
 {
+    
+
```
CJentzsch: please remove this line

PR #131, opened 2015-12-14
delegatecall + call state tests

winsvega:
fix 10mbTransaction s value


winsvega:
hold on there seems to be an issue

libethereum/Account.h
```diff
@@ -213,18 +213,20 @@ class AccountMask
 		m_hasNonce(_all),
 		m_hasCode(_all),
 		m_hasStorage(_all)
-	{}
+	{ m_shouldNotExist = false; }
```
chriseth: Move that to the member itself?

test/TestHelper.cpp
```diff
@@ -287,23 +287,32 @@ int ImportTest::compareStates(State const& _stateExpect, State const& _statePost
 		}
 
 	for (auto const& a: _stateExpect.addresses())
-	{
-		CHECK(_statePost.addressInUse(a.first), TestOutputHelper::testName() +  "Compare States: " << a.first << " missing expected address!");
-		if (_statePost.addressInUse(a.first))
+	{
```
chriseth: Surplus whitespace.

PR #134, opened 2015-12-17
Increased stack offloading constants.

chriseth:


chriseth:
It seems the numbers need to be further increased for macos.
Probably better to make a thorough test on both macos and windows in both debug and release.

PR #136, opened 2015-12-17
Reduce stack size for calls.

chriseth:
This includes three changes:
1. Return EVMSchedule by reference in internal functions
2. Allocate CallParams on the heap
3. Move most of the opcode implementation to its own function

Surprisingly, the third point provided the largest boost, as apparently, memory for temporaries is allocated on the stack and not re-used between switch cases or scopes.


winsvega:
macosx still fails

PR #139, opened 2015-12-22
tests merge

winsvega:
run Homestead and Frontier tests from one test branch


chriseth:
Is this good to be merged?

winsvega:
yes, but better be reviewed by Christoph

test/TestHelper.cpp
```diff
@@ -728,6 +729,13 @@ void executeTests(const string& _name, const string& _testPathAppendix, const bo
 	}
 }
 
+void copyFile(std::string const& _source, std::string const& _destination)
+{
+	std::ifstream  src(_source, std::ios::binary);
```
chriseth: Please remove some of the excess spaces (also next line).

test/libethereum/BlockChainTests.cpp
```diff
@@ -231,6 +234,7 @@ void doBlockchainTests(json_spirit::mValue& _v, bool _fillin)
 				State stateExpect(State::Null);
 				ImportTest::importState(o["expect"].get_obj(), stateExpect, expectStateMap);
 				if (ImportTest::compareStates(stateExpect, testChain.topBlock().state(), expectStateMap, Options::get().checkState ? WhenError::Throw : WhenError::DontThrow))
+					if (Options::get().checkState)
 					cerr << testname << endl;
```
chriseth: Indentation

test/TestHelper.h
```diff
@@ -50,6 +50,8 @@ void mine(BlockHeader& _bi, SealEngineFace* _sealer, bool _verify = true);
 namespace test
 {
 
+const int c_testHomesteadBlock = 900000;
```
CJentzsch: According to our last call, I think that needs to be changed back to 1.000.000 (also update the test accordingly)

test/libethereum/BlockChainTests.cpp
```diff
@@ -34,10 +34,13 @@ using namespace json_spirit;
 using namespace dev;
 using namespace dev::eth;
 
+
 namespace dev {
 
 namespace test {
 
+eth::Network TestBlockChain::sealEngineNetwork = eth::Network::FrontierTest;
```
CJentzsch, line 42: One of my last PRs introduced the ability to specifiy the sealEngine as commandline argument. This would need an update too. Because this would not work anymore with your changes (but may be doesn't have too, so you can remove it.).

CJentzsch, line 42: Also, currently you have `FrontierTest`  and `HomesteadTest` . Which are not enough IMO. Frontier is everthing according to Frontier rules, and Homestead for Homestead rules, but where do you have the switch?
I introduced simple `Tests` where everything before block X is Frontier, and after it is Homestead. I think you should use that.

CJentzsch:
Those changes are ok for now, but long term, you should remove all the tests which have the same result under Homestead and Frontier, and only have a Frontier and Homestead copy of those tests which have different results.

Alos please check that the transition tests https://github.com/ethereum/libethereum/blob/develop/test/libethereum/StateTestsFiller/stTransitionTestFiller.json
use the agreed block X (now 1.000.000)

test/libethereum/BlockChainTests.cpp
winsvega, line 42: I change the network by reading blocknumber in test fillers.
this cmd option  (seal engine) still needed for user defined tests
this paricular line is just defult value for static variable for blockchain tests.

winsvega:
updated

test/libethereum/BlockChainTests.cpp
CJentzsch, line 42: you say you check that blocknumber in the tests, but why do you need to specify the seal engine for each test?

PR #140, opened 2015-12-23
Fix calculateDifficulty from double maxmin

winsvega:
this expression
result difficulty = max ( mindifficulty, calculatedDifficulty)
should be done at the end of all calculations right?
in line 199
o = max<u256>(minimumDifficulty, o + (u256(1) << (periodCount - 2)));
it looks like this
result difficulty = max ( mindifficulty, calculatedDifficulty)
where calculatedDifficulty = max (mindifficulty, calculatedDifficulty2) + x
So when calculatedDifficulty2 < minDifficulty instead of being
resultDifficulty = max ( mindifficulty, calculatedDifficulty2 + x)
it is
resultDifficulty = max ( mindifficulty, mindifficulty + x)

Martin is using  fixed example in his JS client.

### DIFF
--- a/libethashseal/Ethash.cpp
+++ b/libethashseal/Ethash.cpp
@@ -190,11 +190,14 @@ u256 Ethash::calculateDifficulty(BlockHeader const& _bi, BlockHeader const& _par
 	else
 		// Homestead-era difficulty adjustment
 		target = _parent.difficulty() + _parent.difficulty() / 2048 * max<bigint>(1 - (bigint(_bi.timestamp()) - _parent.timestamp()) / 10, -99);
-	u256 o = (u256)max<bigint>(minimumDifficulty, target);
+
+	bigint o = target;
 	unsigned periodCount = unsigned(_parent.number() + 1) / c_expDiffPeriod;
 	if (periodCount > 1)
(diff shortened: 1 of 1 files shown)


libethashseal/Ethash.cpp
```diff
@@ -185,11 +185,14 @@ u256 Ethash::calculateDifficulty(BlockHeader const& _bi, BlockHeader const& _par
...
-	u256 o = (u256)max<bigint>(minimumDifficulty, target);
+
+	bigint o = target;
 	unsigned periodCount = unsigned(_parent.number() + 1) / c_expDiffPeriod;
 	if (periodCount > 1)
-		o = max<u256>(minimumDifficulty, o + (u256(1) << (periodCount - 2)));	// latter will eventually become huge, so ensure it's a bigint.
-	return o;
+		o += (u256(1) << (periodCount - 2));	// latter will eventually become huge, so ensure it's a bigint.
+
+	o = max<bigint>(minimumDifficulty, o);
```
chriseth: I would rather make that a `max<u256>(minDifficulty, o);`

winsvega:
what about comment? 
"// latter will eventually become huge, so ensure it's a bigint."

chriseth:
It is converted to a u256 anyway in the next line. The only difference is that if you use `max<u256>(...)`, you can be sure that the value is larger than `minimumDifficulty`, even in the case of an overflow.

PR #143, opened 2015-12-29
Fix Boost 1.60 build

LefterisJP:
This PR fixes the libethereum build for the new boost 1.60 version.

@winsvega Please have a look at the tests fix. Understand that what you are doing [in boostTest](https://github.com/ethereum/libethereum/blob/develop/test/boostTest.cpp#L113) has to change. We can't be fixing it for each new boost release.

Please don't use internal boost functions.


winsvega:
I feel like those lines with boost filter are not needed at all.

LefterisJP:
All the better then, remove them if they are not needed. They are going to keep breaking the build with each new version of boost if they stay as they are.

PR #145, opened 2016-01-15
Fix sequence number bug.

chriseth:
This bug resulted in incorrect storage access in some situations.
The reason was that when intersecting states, the sequence numbers
were not handled and thus some operations with too low
sequence numbers were used during code generation.

DEPENDS:
{
"solidity":"fix_storage"
}

Fixes solidity#313


chriseth:
Test failures are unrelated.

PR #146, opened 2016-01-18
fixing test check for jenkins

winsvega:


test/libethereum/BlockChain.cpp
```diff
@@ -38,6 +38,7 @@ BOOST_AUTO_TEST_SUITE(BlockChainSuite)
 
 BOOST_AUTO_TEST_CASE(output)
 {
+	TestOutputHelper::initTest();
```
chriseth, line 41: For the next time, it is probably better to create a fixture here.

chriseth:
Gas pricer still fails. Is that an issue of the tests?

winsvega:
I don't think so. windows tests are fine for whatever reason. 
it seems that gaspricer test has a diffrent result depend on a platform for frontier network

CJentzsch:
@winsvega As mentioned before, I would suggest making the GasPricer checks range based checks. Those are effectively floats represented as integer. It would make more sense ff the return value is checked against a range instead a fixed number.

chriseth:
Ah, I did not realise that fact. Christoph is right, we should not compare them at unit precision. This might create a lot of confusion if the tests are run on different platforms.

winsvega:
could we merge this and then open a PR for range checks?

PR #147, opened 2016-01-18
Use bigints for difficulty calculation.

chriseth:


winsvega:
whats the point if return value is 
u256 ?

I see. so if overflow u256 then return max(u256)

PR #152, opened 2016-01-25
Delegatecall and dynamic schedule for EVMJIT.

chriseth:


chfast:
I don't have time to check every detail, but I've checked the idea.

Generally speaking I don't like the direction it is going. Looks like pushing the design of EVM Schedule to JIT is not a good idea. Maybe it looks nice but performance seems to be sacrificed. I liked the idea of dropping cache at some point much better.

Two main problems:
1. I have to manage JitSchedule every single call. And it will almost never change.
2. After Schedule change (e.g. frontier -> homestead) a lots of changed code is left but it will be never used again.

chriseth:
@chfast thanks for taking a look! Do you have an impression about how large the performance sacrifice migth be? I don't think we have any drawback when executing a compiled contract. At compilation time, I would say that the fact whether we put absolute constants or numbers that might vary into llvm ir code does not make a big difference (please correct me if I'm wrong).
The only real performance impact I can see is when copying and converting the schedule. This is probably the equivalent of converting 5 u256 numbers to i256.

I considered using an identifier even at the EVM side and cache the conversion, but I decided that it is not worth the hassle and code complication.

The persistence of the cache is an issue, but I thought it would be very much similar to a version change of the evmjit where the cache is not claened either (as far as I can see). How often did that happen?

chfast:
I have no idea what the performance impact it is. But the Schedule curries too much information and in the same time it is not generic enough to handle future changes. It will be enough to have a "mode" like frontier / homestead / morten, etc.

Cache is not only about disk, all the cache is kept in memory also.

chriseth:
Converted most of the parameters to constants, leaving only a single flag. This might be a  bit ugly, but is should be performing. After sync, my compilation cache has about 180 MB. I think this is fine for now. I created a story to make the jit cache smarter.

evmjit/include/evmjit/JIT.h
```diff
@@ -151,13 +190,19 @@ class JIT
 	/// Ask JIT if the EVM code is ready for execution.
 	/// Returns `true` if the EVM code has been compiled and loaded into memory.
 	/// In this case the code can be executed without overhead.
-	/// \param _codeHash	The Keccak hash of the EVM code.
-	EVMJIT_API static bool isCodeReady(h256 const& _codeHash);
+	/// \param _codeIdentifier	the identifier of the code consisting of a hash of the code and the schedule.
```
LianaHus: you have tab here

PR #156, opened 2016-01-27
remove boost version dependent code

winsvega:
this might be not needed for actual testing. but it makes testeth to crash each time boost lib is updated.

### DIFF
--- a/test/boostTest.cpp
+++ b/test/boostTest.cpp
@@ -107,41 +107,6 @@ int main( int argc, char* argv[] )
 	for (int i = 0; i < argc; i++)
 		parameters.push_back(argv[i]);
 
-	try
-	{
-		framework::init(fake_init_func, argc, argv);
+	unit_test_main(fake_init_func, argc, argv);
 
-		framework::run();
-
-		results_reporter::make_report();
-
-#if BOOST_VERSION >= 106000
-		return results_collector.results(framework::master_test_suite().p_id).result_code();
-#else
-		return runtime_config::no_result_code()
-					? boost::exit_success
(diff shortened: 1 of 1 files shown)


LefterisJP:
It's really good to remove this.

But any idea why the ubuntu tests are failing?

```
Stacktrace

[Exception] - critical check uncleBlockQueue.blockStatus(block2.blockHeader().hash()) == QueueStatus::Bad failed
 == [File] - /var/jenkins_home/sharedspace/libethereum/label/master/libethereum/test/libethereum/BlockChain.cpp
 == [Line] -180
```

[Link](http://52.28.164.97/job/project-test/label=master/1254/testReport/junit/Master.Test.Suite.BlockChainSuite/_var_jenkins_home_sharedspace_libethereum_label_master_libethereum_test_TestHelper/sync/) to failure

chriseth:
@LefterisJP that failure has happened before and is unrelated. We will focus on testing this and the following week and hopefully also fix that problem.

test/boostTest.cpp
```diff
@@ -110,21 +110,6 @@ int main( int argc, char* argv[] )
 	try
 	{
 		framework::init(fake_init_func, argc, argv);
-#if BOOST_VERSION >= 106000
```
chriseth, line 113: What was the purpose of this code if it can be simply removed?

LefterisJP, line 113: Ask @winsvega , the very first usage of boost internal code is by him. I just had to go in and add a new condition each time boost upgraded its minor version.

winsvega, line 113: it is not my code neither. It was here before and it is just appears that it might not even be needed.

PR #158, opened 2016-02-01
fix segfault when typing web3 into the eth console

winsvega:


chriseth:
Are you sure this was the reason? I think you introduced a bug by that change (you skipped one value).

winsvega:
which value?
web3 command works for me with this change.

chriseth:
In the else-case, you first increment `i` and then access `i->second.first`. Before your change, it was accessing `i->second.first` before the increment (keep in mind that it used the post-increment and not the pre-increment operator).

chriseth:
I don't see a semantic change here. If size is zero, then begin should be equal to end and the loop is skipped. On the other hand, if begin is equal to end, size is zero.

winsvega:
what might be the case here? maybe keyword mutable makes this variable vulnerable to changes where it is not needed? 

however if I remove size check it will throw segfault when trying to execute boost add:
ret += i++->second.first;

chriseth:
If the error is really here, then it might be due to heap corruption or a threading issue.

winsvega:
then just checking out size would not be a perfect solution.

PR #159, opened 2016-02-02
stateHomestead on HighS value check

winsvega:


test/libethereum/State.cpp
```diff
@@ -96,6 +96,16 @@ void doStateTests(json_spirit::mValue& _v, bool _fillin)
 
 BOOST_AUTO_TEST_SUITE(StateTests)
 
+BOOST_AUTO_TEST_CASE(stCallDelegateCodesCallCode)
+{
+	dev::test::executeTests("stCallDelegateCodesCallCode", "/StateTests",dev::test::getFolder(__FILE__) + "/StateTestsFiller", dev::test::doStateTests);
```
LianaHus: space

PR #163, opened 2016-02-09
Guard threaded access of the hashrate.

chriseth:
Fixes https://github.com/ethereum/webthree-umbrella/issues/188 and https://github.com/ethereum/webthree-umbrella/issues/113

### DIFF
--- a/libethashseal/EthashClient.cpp
+++ b/libethashseal/EthashClient.cpp
@@ -29,14 +29,14 @@ using namespace p2p;
 EthashClient& dev::eth::asEthashClient(Interface& _c)
 {
 	if (dynamic_cast<Ethash*>(_c.sealEngine()))
-		return reinterpret_cast<EthashClient&>(_c);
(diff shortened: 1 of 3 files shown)


bobsummerwill:
Simple change.   Looks good to me.
Were you able to replicate the issue and verify the fix at runtime, @chriseth?

chriseth:
@bobsummerwill I have seen the issue at some point in the past, I did not verify the fix, but it is obvious that we need a lock here, as `submitExternalHashrate` is called from the json-rpc-thread and `hashrate` is called from the UI thread.

If we have time, we could perhaps disentangle threaded access somehow...

smartbitcoin:
As the talk between  Chris and me , the issue come from the improper casting in EthashClient.cpp. Add a lock wont help this case.

chriseth:
The real fix will be in https://github.com/ethereum/webthree/pull/106

PR #166, opened 2016-02-11
Fix crash in OpenCL miner. Add a check for the maximum allocation siz…

michaelmarks:
…e in addition to the total memory size. Fixes crash on GPUs that could not support a single 1GB allocation


michaelmarks:
Closing request, will work on improved solution version. The current OpenCL miner makes a ~1GB alloc for the full DAG and ignores the max alloc size for the device (CL_DEVICE_MAX_MEM_ALLOC_SIZE). This allocation was crashing some of my Intel GPUs. However my 2GB AMD GPUs reports it only supports a 512MB max alloc, yet the allocation succeeds and mining seems to work. I'm going to investigating getting the old chunking approach to work.

PR #169, opened 2016-02-16
Remove the bad block sentinel mechanism.

chriseth:
Fixes https://github.com/ethereum/webthree-umbrella/issues/215


LianaHus:
it seems you have to remove more :) builds are failing

bobsummerwill:
Woohoo!   Lots of deleted code.   I like it.    Just looking at, and trying to understand the test failures.   @gcolvin - Probably good for you to have a look at these too, so we know what "normal" looks like and "broken" looks like.

bobsummerwill:
@chriseth You just merged.   Does that mean that the test failures were not real issues?   They appear to have disappear from this Conversation following the merge.

UPDATE - Oh no, they just got "folded".   Still visible above.

bobsummerwill:
ie. http://52.28.164.97/job/project-test/label=master/1457/console.

Do you have some comments on this, @winsvega?   Also, what is with the "bad characters" in that output?

chriseth:
@bobsummerwill @gcolvin better to take a look here: http://52.28.164.97/job/project-test/label=master/1457/
So there are two failing tests and they are whisper tests. They have been sporadically failing for a long time. We might disable them until we have more time to tend to whisper again.

bobsummerwill:
Gotcha.   Makes sense.  Added https://github.com/ethereum/webthree-umbrella/issues/220.

PR #170, opened 2016-02-16
update homestead block number in test fillers

winsvega:


test/libethereum/StateTestsFiller/Homestead/stSpecialTestFiller.json
```diff
@@ -266,237 +266,13 @@
         }
     },
 
-    "block504980" : {
```
chriseth, line 269: Why was that removed?

winsvega, line 269: it's a dublicate of 
https://github.com/ethereum/libethereum/blob/develop/test/libethereum/StateTestsFiller/stSpecialTestFiller.json#L269

block 504980 is not of homestead

PR #171, opened 2016-02-17
init state as Idle

yann300:


libethereum/BlockChainSync.cpp
```diff
@@ -137,6 +137,7 @@ BlockChainSync::BlockChainSync(EthereumHost& _host):
 	m_startingBlock(_host.chain().number()),
 	m_lastImportedBlock(m_startingBlock)
 {
+	m_state = SyncState::Idle;
```
chriseth: Why not change the value in the member itself? reassigning it here might be confusing.

yann300:
done

libethereum/BlockChainSync.cpp
```diff
@@ -134,8 +134,9 @@ template<typename T> void mergeInto(std::map<unsigned, std::vector<T>>& _contain
 
 BlockChainSync::BlockChainSync(EthereumHost& _host):
 	m_host(_host),
+	m_state(SyncState::Idle),
 	m_startingBlock(_host.chain().number()),
-	m_lastImportedBlock(m_startingBlock)
+	m_lastImportedBlock(m_startingBlock),
```
chriseth: build error

PR #174, opened 2016-02-19
Split chain params

yann300:
eth takes two arg that can be used to start custom chain.
 --config: json containing network pref, starting state, genesis, mining algo
 --genesis-json:  --genesis json containing only genesis pref
previously providing a simple genesis file leads to a crash

DEPENDS: {
"webthree": "chainParam4"
}

### DIFF
--- a/libethereum/ChainParams.cpp
+++ b/libethereum/ChainParams.cpp
@@ -49,44 +49,67 @@ ChainParams::ChainParams()
 
 ChainParams::ChainParams(std::string const& _json, h256 const& _stateRoot)
 {
+	try
+	{
+		loadConfig(_json, _stateRoot);
+	}
+	catch (...)
+	{
+		cerr << "Bad json file provided. Should be a config file" << endl;
+	}
+}
+
+void ChainParams::loadConfig(std::string const& _json, h256 const& _stateRoot)
+{
 	js::mValue val;
 	json_spirit::read_string(_json, val);
 	js::mObject obj = val.get_obj();
+
 	sealEngineName = obj["sealEngine"].get_str();
-	js::mObject genesis = obj["genesis"].get_obj();
-	js::mObject params = obj["params"].get_obj();
-
 	// params
-	{
-		accountStartNonce = u256(fromBigEndian<u256>(fromHex(params["accountStartNonce"].get_str())));
-		maximumExtraDataSize = u256(fromBigEndian<u256>(fromHex(params["maximumExtraDataSize"].get_str())));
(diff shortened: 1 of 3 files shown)


libethereum/ChainParams.cpp
```diff
@@ -49,44 +49,67 @@ ChainParams::ChainParams()
 
 ChainParams::ChainParams(std::string const& _json, h256 const& _stateRoot)
 {
+	try
+	{
+		loadConfig(_json, _stateRoot);
+	}
+	catch (...)
+	{
+		cerr << "Bad json file provided. Should be a config file" << endl;
```
chriseth, line 58: Please leak the exception - it should not be possible to construct a ChainParams object with an invalid config file. Furthermore, we should not assume that `cerr` is available / read here.

chriseth:
Currently, it is not possible to modify ChainParams (apart from its members being public, but that is a different story). Furthermore, if some exception happens inside load\* (very likely due to invalid json), the ChainParams object is left in an invalid state. What about creating const functions that return a copy of the current object with the specified changes?

libethereum/ChainParams.cpp
```diff
@@ -49,44 +49,67 @@ ChainParams::ChainParams()
...
-	{
-		accountStartNonce = u256(fromBigEndian<u256>(fromHex(params["accountStartNonce"].get_str())));
-		maximumExtraDataSize = u256(fromBigEndian<u256>(fromHex(params["maximumExtraDataSize"].get_str())));
-		tieBreakingGas = params.count("tieBreakingGas") ? params["tieBreakingGas"].get_bool() : true;
-		blockReward = u256(fromBigEndian<u256>(fromHex(params["blockReward"].get_str())));
-		for (auto i: params)
-			if (i.first != "accountStartNonce" && i.first != "maximumExtraDataSize" && i.first != "blockReward" && i.first != "tieBreakingGas")
-				otherParams[i.first] = i.second.get_str();
-	}
+void ChainParams::loadGenesisState(std::string const& _json, std::unordered_map<Address, PrecompiledContract> const& _precompiled)
```
chriseth, line 85: No `std::`

libethereum/ChainParams.cpp
```diff
@@ -49,44 +49,67 @@ ChainParams::ChainParams()
...
+{
+	precompiled = _precompiled;
+	genesisState = jsonToAccountMap(_json, nullptr, &precompiled);
```
chriseth, line 88: If this is not done for `loadGenesis`, are the account balances of the genesis block correctly initialised?

chriseth:
Please make sure to test whether you can sync to frontier when supplying the frontier genesis json.

PR #175, opened 2016-02-22
Split chain params

yann300:
eth takes two arg that can be used to start custom chain.
--config: json containing network pref, starting state, genesis, mining algo
--genesis-json: --genesis json containing only genesis pref
previously providing a simple genesis file leads to a crash

DEPENDS: {
"webthree": "chainParam4"
}

### DIFF
--- a/libethereum/ChainParams.cpp
+++ b/libethereum/ChainParams.cpp
@@ -34,6 +34,42 @@ using namespace dev;
 using namespace eth;
 namespace js = json_spirit;
 
+static ChainParams setGenesis(string const& _json, h256 const& _stateRoot, ChainParams const& _org)
+{
+	ChainParams cp = ChainParams(_org);
+
+	js::mValue val;
+	json_spirit::read_string(_json, val);
+	js::mObject genesis = val.get_obj();
+
+	cp.stateRoot = _stateRoot;
+	cp.parentHash = h256(genesis["parentHash"].get_str());
+	cp.author = genesis.count("coinbase") ? h160(genesis["coinbase"].get_str()) : h160(genesis["author"].get_str());
+	cp.difficulty = genesis.count("difficulty") ? u256(fromBigEndian<u256>(fromHex(genesis["difficulty"].get_str()))) : 0;
+	cp.gasLimit = u256(fromBigEndian<u256>(fromHex(genesis["gasLimit"].get_str())));
+	cp.gasUsed = genesis.count("gasUsed") ? u256(fromBigEndian<u256>(fromHex(genesis["gasUsed"].get_str()))) : 0;
+	cp.timestamp = u256(fromBigEndian<u256>(fromHex(genesis["timestamp"].get_str())));
+	cp.extraData = bytes(fromHex(genesis["extraData"].get_str()));
+
+	// magic code for handling ethash stuff:
+	if ((genesis.count("mixhash") || genesis.count("mixHash")) && genesis.count("nonce"))
+	{
+		h256 mixHash(genesis[genesis.count("mixhash") ? "mixhash" : "mixHash"].get_str());
+		h64 nonce(genesis["nonce"].get_str());
+		cp.sealFields = 2;
+		cp.sealRLP = rlp(mixHash) + rlp(nonce);
+	}
+	return cp;
+}
+
+static ChainParams setGenesisState(string const& _json, unordered_map<Address, PrecompiledContract> const& _precompiled, ChainParams const& _org)
+{
+	ChainParams cp = ChainParams(_org);
+	cp.precompiled = _precompiled;
+	cp.genesisState = jsonToAccountMap(_json, nullptr, &cp.precompiled);
+	return cp;
+}
+
 ChainParams::ChainParams()
 {
 	genesisState[Address(1)] = Account(0, 1);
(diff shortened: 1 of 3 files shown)


libethereum/ChainParams.cpp
```diff
@@ -47,47 +83,75 @@ ChainParams::ChainParams()
...
+{
+	loadConfig(_json, _stateRoot);
+}
+
+ChainParams::ChainParams(ChainParams const& _org)
+{
+	copy(_org);
+}
+
+void ChainParams::copy(ChainParams const& _org)
```
chriseth: Why not put that directly in the copy constructor? You could use member initializer for most of the fields. Or actually use `=default`?

libethereum/ChainParams.cpp
```diff
@@ -47,47 +83,75 @@ ChainParams::ChainParams()
...
+	gasUsed = _org.gasUsed;
+	timestamp = _org.timestamp;
+	extraData = _org.extraData;
+	sealFields = _org.sealFields;
+	sealRLP = _org.sealRLP;
+}
+
+ChainParams ChainParams::loadConfig(string const& _json, h256 const& _stateRoot) const
+{
+	ChainParams cp = ChainParams(*this);
```
chriseth: `ChainParams cp(*this)`?

libethereum/ChainParams.cpp
```diff
@@ -34,6 +34,42 @@ using namespace dev;
 using namespace eth;
 namespace js = json_spirit;
 
+static ChainParams setGenesis(string const& _json, h256 const& _stateRoot, ChainParams const& _org)
```
chriseth: You could make that a private function

libethereum/ChainParams.cpp
```diff
@@ -47,47 +47,110 @@ ChainParams::ChainParams()
...
 	json_spirit::read_string(_json, val);
-	js::mObject obj = val.get_obj();
-	sealEngineName = obj["sealEngine"].get_str();
-	js::mObject genesis = obj["genesis"].get_obj();
-	js::mObject params = obj["params"].get_obj();
-
-	// params
+	js::mObject genesis = val.get_obj();
+
+	cp.stateRoot = _stateRoot;
```
chriseth: Please add 'stateRoot = _stateRoot ? _stateRoot : cp.calculateStateRoot();' here, but only after having set the genesis state.

libethereum/ChainParams.cpp
```diff
@@ -34,60 +34,110 @@ using namespace dev;
...
+	js::mObject params = obj["params"].get_obj();
+	cp.accountStartNonce = u256(fromBigEndian<u256>(fromHex(params["accountStartNonce"].get_str())));
+	cp.maximumExtraDataSize = u256(fromBigEndian<u256>(fromHex(params["maximumExtraDataSize"].get_str())));
+	cp.tieBreakingGas = params.count("tieBreakingGas") ? params["tieBreakingGas"].get_bool() : true;
+	cp.blockReward = u256(fromBigEndian<u256>(fromHex(params["blockReward"].get_str())));
+	for (auto i: params)
+		if (i.first != "accountStartNonce" && i.first != "maximumExtraDataSize" && i.first != "blockReward" && i.first != "tieBreakingGas")
+			cp.otherParams[i.first] = i.second.get_str();
 	// genesis
+	if (_importGenesis)
```
chriseth: I think it is better to import it and overwrite it when needed instead of having this boolean flag.

libethereum/ChainParams.h
```diff
@@ -36,7 +36,8 @@ class SealEngineFace;
 
 struct ChainParams: public ChainOperationParams
 {
-	ChainParams();
+	ChainParams() = default;
+	ChainParams(ChainParams const& _org);
```
chriseth: What about using `= default` also here?

libethereum/ChainParams.cpp
```diff
@@ -34,60 +34,107 @@ using namespace dev;
...
-		}
-	}
+	stringstream strGenesis;
+	string genesisStr = json_spirit::write_string(obj["genesis"], false);
+	setGenesis(genesisStr, _stateRoot, cp);
+	// genesis state
+	stringstream strGenesisState;
+	string genesisStateStr = json_spirit::write_string(obj["accounts"], false);
+	setGenesisState(genesisStateStr, unordered_map<Address, PrecompiledContract>(), cp);
+	return cp;
```
chriseth: I think you should set the state root only here (using the line with the ternary operator), because it depends on the genesis state.

libethereum/ChainParams.cpp
```diff
@@ -34,60 +34,107 @@ using namespace dev;
...
+}
+
+ChainParams ChainParams::loadGenesisState(string const& _json, unordered_map<Address, PrecompiledContract> const& _precompiled) const
+{
+	return setGenesisState(_json, _precompiled, *this);
```
chriseth: Why do we need the private function? Can't we just move its implementation here? (same thing applies for the other private function)

libethereum/ChainParams.cpp
```diff
@@ -34,60 +34,107 @@ using namespace dev;
...
-		blockReward = u256(fromBigEndian<u256>(fromHex(params["blockReward"].get_str())));
-		for (auto i: params)
-			if (i.first != "accountStartNonce" && i.first != "maximumExtraDataSize" && i.first != "blockReward" && i.first != "tieBreakingGas")
-				otherParams[i.first] = i.second.get_str();
+		h256 mixHash(genesis[genesis.count("mixhash") ? "mixhash" : "mixHash"].get_str());
+		h64 nonce(genesis["nonce"].get_str());
+		cp.sealFields = 2;
+		cp.sealRLP = rlp(mixHash) + rlp(nonce);
 	}
+	cp.stateRoot = _stateRoot ? _stateRoot : cp.calculateStateRoot();
```
chriseth: Hm, we have not yet set / not modified the genesis state here. This should be moved to after the call of setGenesisState or inside it.

libethereum/ChainParams.cpp
```diff
@@ -34,60 +34,72 @@ using namespace dev;
...
-	genesisState[Address(1)] = Account(0, 1);
-	genesisState[Address(2)] = Account(0, 1);
-	genesisState[Address(3)] = Account(0, 1);
-	genesisState[Address(4)] = Account(0, 1);
-	// Setup default precompiled contracts as equal to genesis of Frontier.
-	precompiled.insert(make_pair(Address(1), PrecompiledContract(3000, 0, PrecompiledRegistrar::executor("ecrecover"))));
-	precompiled.insert(make_pair(Address(2), PrecompiledContract(60, 12, PrecompiledRegistrar::executor("sha256"))));
-	precompiled.insert(make_pair(Address(3), PrecompiledContract(600, 120, PrecompiledRegistrar::executor("ripemd160"))));
-	precompiled.insert(make_pair(Address(4), PrecompiledContract(15, 3, PrecompiledRegistrar::executor("identity"))));
+	loadConfig(_json, _stateRoot);
```
chriseth: `*this = loadConfig(...)`

libethereum/ChainParams.cpp
```diff
@@ -34,60 +34,72 @@ using namespace dev;
...
+	cp = cp.loadGenesisState(genesisStateStr, unordered_map<Address, PrecompiledContract>());
+	return cp;
+}
+
+ChainParams ChainParams::loadGenesisState(string const& _json, unordered_map<Address, PrecompiledContract> const& _precompiled) const
+{
+	ChainParams cp(_org);
+	cp.precompiled = _precompiled;
+	cp.genesisState = jsonToAccountMap(_json, nullptr, &cp.precompiled);
+	cp.stateRoot = cp.calculateStateRoot();
```
chriseth: Only if it is not explicitly given - I think you have to add it as an optional parameter here

libethereum/ChainParams.cpp
```diff
@@ -34,60 +34,72 @@ using namespace dev;
...
+	cp = cp.loadGenesis(genesisStr, _stateRoot);
+	// genesis state
+	string genesisStateStr = json_spirit::write_string(obj["accounts"], false);
+	cp = cp.loadGenesisState(genesisStateStr, unordered_map<Address, PrecompiledContract>());
+	return cp;
+}
+
+ChainParams ChainParams::loadGenesisState(string const& _json, unordered_map<Address, PrecompiledContract> const& _precompiled) const
+{
+	ChainParams cp(_org);
```
chriseth: `_org` -> `*this`

libethereum/ChainParams.cpp
```diff
@@ -34,60 +34,72 @@ using namespace dev;
...
+	cp.precompiled = _precompiled;
+	cp.genesisState = jsonToAccountMap(_json, nullptr, &cp.precompiled);
+	cp.stateRoot = cp.calculateStateRoot();
+	return cp;
+}
+
+ChainParams ChainParams::loadGenesis(string const& _json, h256 const& _stateRoot) const
+{
+	ChainParams cp(_org);
```
chriseth: `_org` -> `*this`

PR #181, opened 2016-02-25
Fix OS X build break versus LLVM HEAD

bobsummerwill:
This looks like it has been caused by cleanup of iterator semantics in LLVM, which exposed an implicit conversation which isn't allowed anymore.

### DIFF
--- a/evmjit/libevmjit/Optimizer.cpp
+++ b/evmjit/libevmjit/Optimizer.cpp
@@ -39,7 +39,8 @@ char LongJmpEliminationPass::ID = 0;
 
 bool LongJmpEliminationPass::runOnFunction(llvm::Function& _func)
 {
-	if (&_func != _func.getParent()->begin())
+	auto iter = _func.getParent()->begin();
+	if (&_func != &(*iter))
 		return false;
 
 	auto& mainFunc = _func;


bobsummerwill:
CC @chfast

evmjit/libevmjit/Optimizer.cpp
```diff
@@ -39,7 +39,8 @@ char LongJmpEliminationPass::ID = 0;
 
 bool LongJmpEliminationPass::runOnFunction(llvm::Function& _func)
 {
-	if (&_func != _func.getParent()->begin())
+	auto iter = _func.getParent()->begin();
+	if (&_func != &(*iter))
```
chfast, line 43: This code is not enabled by default so you should test it first.

bobsummerwill:
Any recommendations on the best way to do that, @chfast?   Thanks!

bobsummerwill:
FYI ... I think I'm likely to be in a position to get OS X working with the LLVM 3.7.1 versioned release soon, which should settle these issues down.

https://github.com/ethereum/webthree-umbrella/issues/201

chfast:
Yes, sorry for late response.

I've checked the code. It should be enough to run all tests with following env variables set:
`EVMJIT="-O -cache=0"`.

In general, you can pass options by `EVMJIT` variable to any process running jit. You can try with `EVMJIT="-help"` or `EVMJIT="-version"`. Some of the options are from LLVM.

I don't expect there to be any issues. If all test pass you can merge it.

bobsummerwill:
@chfast Do you mean?

```
EVMJIT="-O -cache=0" webthree-helpers/scripts/ethtests.sh libethereum --umbrella
```

If so, those are working fine.

Or are you talking about the "tests" repo?   What is the command-line invocation for that?   CC @winsvega.

chfast:
`EVMJIT="-O -cache=0" testeth --vm jit`

chriseth:
running that now

chriseth:
Run was fine :-)

bobsummerwill:
Thanks, chriseth.    What exactly did you run?   And what's your command-line for doing that?

Just still catching up on how the tests are put together.

PR #182, opened 2016-02-26
Integrate KeyCLI in eth

yann300:
All options used in ethkey should be available using eth options: 
eth account
eth wallet

DEPENDS: {
"libethereum": "keymanager"
}


chriseth:
We decided to make it not depend on ethkey

PR #183, opened 2016-02-26
OS X - Correct default path to LLVM HEAD install's cmake files.

bobsummerwill:
This path changed at some point in the past month or two.
In the absence of this path being correct, cmake invocations have to have this path explicitly passed (and in fact our instructions do, which I never knew the reason for).    So we shouldn't need that.    The default here should just be correct :-)

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -26,7 +26,7 @@ if (EVMJIT)
 		if (DEFINED MSVC)
 			set(LLVM_DIR "${ETH_CMAKE_DIR}/extdep/install/windows/x64/share/llvm/cmake")
 		elseif (DEFINED APPLE)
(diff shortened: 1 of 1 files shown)


bobsummerwill:
NOTE - This will show as broken because the automated build process is NOT touching the Homebrew environment at all, so this machine will just have whatever was installed on when it was set up.

That actually means that it isn't an accurate representation of what end-users see.

We should probably AT LEAST do `brew update && brew upgrade` on it per-cycle.    Short of that, we should do a manual update _now_ at the same time as submitting this change.

Added as https://github.com/ethereum/webthree-umbrella/issues/276

bobsummerwill:
Please can whoever does this also `brew install llvm37` at the same time, so we have both LLVM HEAD and 3.7.1 available on the machine?   Thanks!

PR #187, opened 2016-02-29
Force recalculation of state root if genesis accounst changed.

chriseth:


chriseth:
Tests are known to fail, will fix later.

PR #193, opened 2016-03-01
Correctly set account start nonce upon chainparams creation.

chriseth:
Fixes morden consensus.

### DIFF
--- a/libethereum/Account.cpp
+++ b/libethereum/Account.cpp
@@ -44,7 +44,7 @@ uint64_t toUnsigned(js::mValue const& _v)
 	}
 }
 
-AccountMap dev::eth::jsonToAccountMap(std::string const& _json, AccountMaskMap* o_mask, PrecompiledContractMap* o_precompiled)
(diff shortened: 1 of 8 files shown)


chriseth:
It seems this is not yet the full fix, cannot sync beyond 323, I would assume 324 contains the first transaction from an account that is not part of the genesis block and that account has an invalid start nonce (morden has a start nonce of != 0).

chriseth:
Oh and I have seen a commit in an ethcore repo that was not propagated upwards that changed a default constructor - that might be the fix.

bobsummerwill:
Just for my understanding, @chriseth, so the issue here is that we were implicitly dependent on the fact that the nonce for the genesis block was 0, which is true for the mainnet, but false for Morden.

Is this a consensus bug something new which only affects post-Homestead blocks?

What would that nonce be for other arbitrary private networks?

chriseth:
@bobsummerwill the nonce is only different for morden to avoid replay attacks between the networks. As far as I can tell, morden never worked in c++, a lot of changes connected to the start nonce were just not implemented.

bobsummerwill:
Gotcha.   Thanks for the info.

I guess, thinking back, that I have always used geth on the handful of occasions where I have used Morden (including on the Gear S2 smartwatch!)

PR #198, opened 2016-03-04
debug block queue status on sync test

winsvega:
test

### DIFF
--- a/libethereum/BlockQueue.cpp
+++ b/libethereum/BlockQueue.cpp
@@ -570,3 +570,9 @@ bool BlockQueue::isActive() const
 				return false;
 	return true;
 }
+
+std::ostream& dev::eth::operator << (std::ostream& os, const QueueStatus& obj)
+{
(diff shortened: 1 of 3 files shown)


libethereum/BlockQueue.h
```diff
@@ -65,6 +65,8 @@ enum class QueueStatus
 	Unknown
 };
 
+std::ostream& operator << (std::ostream& os, const QueueStatus& obj);
```
chriseth: `operator<<` and `QueueStatus const&`

libethereum/BlockQueue.cpp
```diff
@@ -570,3 +570,9 @@ bool BlockQueue::isActive() const
 				return false;
 	return true;
 }
+
+std::ostream& dev::eth::operator << (std::ostream& os, const QueueStatus& obj)
```
chriseth: `operator<<` and `QueueStatus const&`

winsvega:
lets add this PR and see if it throws an error in this test on some build again

PR #200, opened 2016-03-07
Save address together with the key

yann300:
related to https://github.com/ethereum/webthree-umbrella/issues/297
step: store addresses with bare keys

DEPENDS: {
"libweb3core": "keyManager2"
}


libethcore/KeyManager.cpp
```diff
@@ -103,6 +103,12 @@ bool KeyManager::load(string const& _pass)
 						m_addrLookup[addr] = uuid;
 						m_uuidLookup[uuid] = addr;
 						m_keyInfo[addr] = KeyInfo(h256(i[2]), string(i[3]), i.itemCount() > 4 ? string(i[4]) : "");
+						SecretStore::EncryptedKey key = m_store.key(uuid);
+						if (!key.encryptedKey.empty() && key.address == ZeroAddress)
+						{
+							key.address = addr;
+							m_store.saveKey(uuid, key);
```
chriseth: Try to call that only once at the end.

PR #202, opened 2016-03-09
Sync stability fixes

arkpar:
- Fixed assertion failure on mismatched block
- Better fork recovery

Ideas for future work:
- Detect and disable peers which are on forked beyond repair (> 1000 blocks)


chriseth:
Continued in https://github.com/ethereum/libethereum/pull/204

PR #208, opened 2016-03-15
Revert bad block

yann300:
- revert bad block

### DIFF
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -456,19 +456,10 @@ void BlockChainSync::onPeerBlockHeaders(std::shared_ptr<EthereumPeer> _peer, RLP
 				if ((prevBlock && prevBlock->hash != info.parentHash()) || (blockNumber == m_lastImportedBlock + 1 && info.parentHash() != m_lastImportedBlockHash))
 				{
 					// mismatching parent id, delete the previous block and don't add this one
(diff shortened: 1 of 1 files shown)


libethereum/BlockChainSync.cpp
```diff
@@ -463,6 +463,11 @@ void BlockChainSync::onPeerBlockHeaders(std::shared_ptr<EthereumPeer> _peer, RLP
 					m_downloadingHeaders.erase(blockNumber - 1);
 					removeItem(m_headers, blockNumber - 1);
 					removeItem(m_bodies, blockNumber - 1);
+					m_lastImportedBlock--;
+					m_highestBlock = m_lastImportedBlock;
```
chriseth: Hm, not totally sure about this change. We do not have `m_lastImportedBlock == blockNumber - 1` in all cases in this if, furthemore, should we actually delete a block that we already imported? Can you check back with arkadiy, perhaps?

arkpar: Right, better do that under the `blockNumber == m_lastImportedBlock + 1 && info.parentHash() != m_lastImportedBlockHash` only and get `m_lastImportedBlockHash` from the block chain or block queue. Better option would be to restart sync here and set `m_haveCommonBlock` to false.
Also resetting `m_highestBlock` does not make much sense. It is used for progress reporting only.

PR #209, opened 2016-03-18
Fix Windows warnings

bobsummerwill:


chriseth:
Are the build errors related?

bobsummerwill:
Aha - yes.   Looks like I over-stripped some defines, and didn't retest on OS X and Linux.   Will fix that up now.

PR #210, opened 2016-03-21
remove libevmasm liblllc lllc

winsvega:


chriseth:
I think you still have to modify the libethereum/CMakeLists file

PR #214, opened 2016-03-23
Move lll.

chriseth:
DEPENDS: {
"webthree-helpers": "movelll",
"solidity": "movesol",
"alethzero": "movesol"
}


test/TestHelper.cpp
```diff
@@ -544,25 +547,28 @@ bytes importData(json_spirit::mObject const& _o)
...
+	FILE *fp = popen(cmd.c_str(), "r");
 	if (fp == NULL)
 		BOOST_ERROR("Failed to run lllc");
-	if (fgets(input, sizeof(input)-1, fp) == NULL)
+	if (fgets(input, sizeof(input) - 1, fp) == NULL)
 		BOOST_ERROR("Reading empty file for lllc");
 	pclose(fp);
 
-	boost::filesystem::remove_all(path);
+	boost::filesystem::remove(path);
```
winsvega, line 567: just remove wont work if it's a directory with files

chriseth, line 567: But why should it be a directory?

winsvega, line 567: because it is

```
556 boost::filesystem::path path(boost::filesystem::temp_directory_path() / boost::filesystem::unique_path());
```

PR #216, opened 2016-03-26
Moved LLVM-specific CMake content from the top-level CMakeLists.txt file

bobsummerwill:
Moved LLVM-specific CMake content from the top-level CMakeLists.txt file for libethereum down into the CMake file for the evmjit module.

Also changed the setting of LLVM_DIR on OS X and Windows from only happening if already undefined to being an unconditional override of any existing settings.    We always want to use the CMake files in those locations and providing the ability for an end-user to override that is a malfeature.   This change is to address a specific break which I am seeing on my MacBook right now, where LLVM_DIR was not undefined.   It was incorrectly defined, and that broke the build.    That won't be possible anymore.     It looks like it was being set to LLVM_DIR-NOT-FOUND by some other element within the CMake process, and that was breaking things here.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -22,19 +22,7 @@ include(EthOptions)
 configure_project(CPUID CURL VMTRACE ROCKSDB OLYMPIC PARANOID TESTS ETHASHCL EVMJIT FATDB)
 
 if (EVMJIT)
-	if (NOT DEFINED LLVM_DIR)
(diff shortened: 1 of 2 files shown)


bobsummerwill:
I'm just going to pull the trigger on this.    The OS X test failures are a [known issue](https://github.com/ethereum/webthree-umbrella/issues/341) which is unrelated to the changes.

PR #219, opened 2016-03-30
fix OOG testFillers

winsvega:
connects to https://github.com/ethereum/webthree-umbrella/issues/407


chriseth:
Please give a bit more details about this change. Why did the number of lines double? What is auto-generated, what is manual and why did you do the changes?

winsvega:
ok so. first of all. all of the test fillers are manual. auto-generated tests are stored in the test repo. 
in this PR I've changed incorrect MSTORE instructions and recalibrated gas parameters for OOG tests. so those test fillers would produce situation where OOG exception would occure before or after calling a DELEGATE/CALL/CALLCODE instruction in certain subcall depth levels

PR #223, opened 2016-04-02
__INTEL_COMPILER

ghost:
Error #396


bobsummerwill:
Found this relevant doc:   http://www.boost.org/doc/libs/1_56_0/libs/multiprecision/doc/html/boost_multiprecision/tut/conversions.html

ethminer/MinerAux.h
```diff
@@ -464,7 +464,11 @@ class MinerCLI
 			PhoneHome rpc(client);
 			try
 			{
+#ifdef __INTEL_COMPILER
+				unsigned ranking = rpc.report_benchmark(platformInfo, (int)innerMean.convert_to<int>());
```
bobsummerwill: Would this work without the C-style cast?
So in all of these cases (and in the other PR too), we would ditch the #ifdef, and just switch the C-cast to the convert_to<> unconditionally, so this particular one, for example would go from ...

`unsigned ranking = rpc.report_benchmark(platformInfo, (int)innerMean);`

to ...

`unsigned ranking = rpc.report_benchmark(platformInfo, innerMean.convert_to<int>());`

So we wouldn't just make that change for the Intel compiler.   We would do it for all compilers.    I don't know why we are doing any C-style casts in the first place, to be honest.   It is not good style.

ghost:
reviewed libethereum #defines:

```
libethereum\ethminer\MinerAux.h: removed defines
libethereum\libethashseal\Ethash.cpp: removed defines
libethereum\libethashseal\EthashAux.cpp: removed defines
libethereum\libethcore\ABI.h: removed defines
libethereum\libethcore\Common.cpp: removed defines
libethereum\libethcore\CommonJS.cpp: removed defines
libethereum\libethcore\ICAP.cpp: ifdef __INTEL_COMPILER: template function toBase36
libethereum\libethcore\Precompiled.cpp: removed defines
libethereum\libethereum\BlockChain.cpp: removed defines
libethereum\libethereum\BlockChainSync.cpp: removed defines
libethereum\libethereum\BlockQueue.cpp: removed defines
libethereum\libethereum\ClientBase.cpp: removed defines
libethereum\libethereum\EthereumPeer.cpp: removed defines
libethereum\libethereum\LogFilter.h: ifdef __INTEL_COMPILER: disable warning 1098
libethereum\libevm\ExtVMFace.h: removed defines
libethereum\libevm\VM.cpp: removed defines
```

libethcore/ICAP.cpp
```diff
@@ -111,7 +111,19 @@ std::string ICAP::encoded() const
 {
 	if (m_type == Direct)
 	{
+#ifdef __INTEL_COMPILER
```
bobsummerwill: So what is happening here?   Why is Intel needing an entirely different code-path?

libethcore/Precompiled.cpp
```diff
@@ -57,8 +57,7 @@ ETH_REGISTER_PRECOMPILED(ecrecover)(bytesConstRef _in, bytesRef _out)
 	u256 v = (u256)in.v;
 	if (v >= 27 && v <= 28)
 	{
-		SignatureStruct sig(in.r, in.s, (byte)((int)v - 27));
-		if (sig.isValid())
```
bobsummerwill, line 61: Conditional lost accidentally here?   The isValid()

libethereum/LogFilter.h
```diff
@@ -26,6 +26,10 @@
 #include <libethcore/Common.h>
 #include "TransactionReceipt.h"
 
+#ifdef __INTEL_COMPILER
+#pragma warning(disable:1098)
```
bobsummerwill: Please can you add a comment here on the RHS of the #pragma line, with the warning message being suppressed?  Thanks!

PR #224, opened 2016-04-07
template workaround

ghost:


chriseth:
Thanks for taking the effort to get this compiling on intel, @mancoast !
I fear that the codebase is plastered with conversions like this. I would really prefer if we could keep explicit conversions like `unsigned(u256(7))`. Perhaps there is some way to introduce a conversion operator?

ghost:
I agree, a conversion function would be ideal, but that would only be a bandaid for the real issue. Lossy conversions with implicit casts. I think It is a true GCCism to enable this type of syntax. What are your thoughts?

chriseth:
These are not lossy conversions with implicit casts, they are lossy conversions with explicit casts and well-defined. Actually I think the root cause is the following. Citing from http://www.boost.org/doc/libs/1_57_0/libs/multiprecision/doc/html/boost_multiprecision/tut/conversions.html:

```
A number can be converted to any built in type, via an explicit conversion operator: this functionality is only available on compilers supporting C++11's explicit conversion syntax.

mpz_int z(2);
int i = z;                     // Error, implicit conversion not allowed.
int j = static_cast<int>(z);   // OK explicit conversion.
```

And it seems that the intel compiler supports c++11 explicit conversion only starting from version 13: https://software.intel.com/en-us/articles/c0x-features-supported-by-intel-c-compiler

bobsummerwill:
Which version of the Intel compiler are you using, @mancoast?
Is updating a possibility?    Or if you are already on a "new enough" version, maybe some extra command-line setting is required to enable this?

ghost:
CC=icc CXX=icpc CFLAGS=-mmic CXXFLAGS=-std=c++11 -mmic LDFLAGS=-mmic
/opt/intel/parallel_studio_xe_2016.2.062/psxevars.sh

bobsummerwill:
Hey @mancoast,

Sorry - need some more context on your comment above.     Are you saying that is the parameters which you currently ARE building with?    Which Intel compiler version are you using?

Any feedback on @chriseth's investigations above and explicit version support?

If there is some way we can NOT ugly up the code, that would obviously be preferable.    Are you using the latest-and-greatest with all the right settings and it STILL isn't supported?

ghost:
Greetings gents!
Apologies for the misunderstandings earlier. I blatantly had confused implicit casts with explicit casts. Regarding the Intel platform version, I run the latest stack and always apply updates ASAP.  The flags used for x86 compilation was simply "-std=c++11".  k1om compilation is a whole other animal that uses the flags "-mmic".  all libraries have been recompiled with k1om format, but unfortunately the c++ library packaged with the Xeon phi are 4.7.0, not 4.8.0. This has been a long battle to get the ethereum code ase to compile using 4.7.0, and it nearly ready for baseline measurements. Then the real fun begins with vtune.

chriseth:
@mancoast hm, that is really strange. By any chance, did you try whether `--std=c++13` makes any difference? Furthermore, do these changes enable compiling _everything_? There are several places that use similar explicit casts like in VM.cpp, and they should fail in the same way.

ghost:
There are many many more uses of the casts that won't compile. I have disabled a vast majority of options including lll and vm. I can work out the remaining casts, but there was a boost internal error using the "spirit" extension within lll. I know I can get vm to completely compile with addition "convert_to" usage, but I did not take the time to resolve the boost "spirit" error encountered during lll compilation.  I will attempt compilation using c++13 for x64 architecture. This will take some time to reset the build from k1om back to x64.

bobsummerwill:
Thanks for your ongoing work here, @mancoast.

Did either of you work out if the Intel compiler WILL add this implicit conversion support at some future point?   Or just in some version of the Intel compiler which is newer than you are currently using @mancoast?

Barring the `json_spirit` compilation errors and other "not everything working" constraints, is it plausible that we just resolve ALL of the implicit conversions (even if there are hundreds of them), and get to a point where we don't need the ifdefs?

Or are we looking at a missing feature in the Intel compiler which really needs to be fixed in the compiler itself?    Is there an issue tracking system for the compiler which we could find or log an issue on this in?

ghost:
Hello @bobsummerwill @chriseth,
I attempted to use --std=c++13, but the argument was unrecognized by icpc.
How should we proceed?
Thanks,
coast

bobsummerwill:
I don't quite know where you got `-std=c++13` from, @chriseth :-)
There is no C++ 13 standard!    And I don't see any such non-standard option in the Intel compiler docs.
https://software.intel.com/en-us/node/523124

@mancoast I still don't understand which exact Intel compiler version you are using yourself.   Is there a --version you could call on your icc executable?

From one line above I think you are using "Intel® Parallel Studio XE 2016 Update 2 Release", which I assume is bundled with Intel compiler 15.x?

bobsummerwill:
"but unfortunately the c++ library packaged with the Xeon phi are 4.7.0, not 4.8.0"

For my Tizen cross-builds, the Gear S2 smartwatch (Tizen 2.3.1) ships with the GLIBC from 4.6.   It is a nightmare.   I'm looking at --sysroot for that, because dropping back to such an old GCC is completely impossible, so I'm need to try to use a newer GCC with an older GLIBC.

bobsummerwill:
No ... from the other issue now, I see ... Intel Compiler 16.

So yeah, if that doesn't work, with `-std=c++11` then I don't think there is anything nicer we can do, is there?

ghost:
Server is running these Intel versions:

```
root@localhost libethereum-buildbot]# icc -V
Intel(R) C Intel(R) 64 Compiler for applications running on Intel(R) 64, Version 16.0.2.181 Build 20160204
root@localhost libethereum-buildbot]# icpc -V
Intel(R) C++ Intel(R) 64 Compiler for applications running on Intel(R) 64, Version 16.0.2.181 Build 20160204
```

I am currently building a new VM for intel compiler 2017 beta.
Will try to work a bug report into their helpdesk to get this thing resolve on the Intel side.

For now, I am satisfied, and the ethminer has found multiple accepted solutions, so its functional.
TODO: future proof resolution

bobsummerwill:
Thanks for the update, @mancoast.
I've mailed the Boost developers mailing list too.

ghost:
confirmed build using boost_1_60_0

PR #226, opened 2016-04-12
opendb test on windows

winsvega:
need to check on windows


chriseth:
Hm, I though the assumption was not that there is another instance running but rather that we do not close the database properly.

winsvega:
it could happen in both scenarious

chriseth:
Please report as soon as this works on windows.

PR #227, opened 2016-04-12
Added ifdefs around pre-processor symbols in the template for ConfigInfo.h.in

bobsummerwill:
I will commit this change as soon as Jenkins shows green, because it will then mean that the following PRs to eliminate this file can be treated atomically.


bobsummerwill:
This is one of the commits for https://github.com/ethereum/webthree-umbrella/issues/456

PR #228, opened 2016-04-13
Eliminated ConfigInfo.h includes from libethereum.

bobsummerwill:
Also removed ETH_TRUE defines, and added various missing "endif" comments, for easier searching of start/end of pre-processing sections.


ethminer/MinerAux.h
```diff
@@ -42,18 +42,23 @@
 #include <libethashseal/EthashGPUMiner.h>
 #include <libethashseal/EthashCPUMiner.h>
 #include <libethashseal/Ethash.h>
-#if ETH_ETHASHCL || !ETH_TRUE
+
+#if ETH_ETHASHCL
```
chfast, line 46: Nice. I like this style!

PR #231, opened 2016-04-13
enable bcBasicInsert

winsvega:


chriseth:
@winsvega you ran the tests locally on windows and they worked?

bobsummerwill:
Looks good, as long as they actually RUN :-)

BTW - Do we have an issue tracking the fact that we should get the Windows tests working in Jenkins?   If not, we really should.

winsvega:
"you ran the tests locally on windows"
yes

@bobsummerwill 
https://github.com/ethereum/webthree-umbrella/issues/335

PR #233, opened 2016-04-14
Please don't ask me why. This partly reverts c1d0429ce604.

chriseth:
This at least changes the values of the `EVMJIT_DLLS` variables in some contexts (verified by using cmake's `message').


chriseth:
I guess the `PARENT_SCOPE` option could be the reason: By moving it down one level, the parent scope changes.

bobsummerwill:
magic wand

PR #236, opened 2016-04-20
vmTrace: add fields required by remix

yann300:


libethereum/Executive.cpp
```diff
@@ -36,7 +37,7 @@ const char* VMTraceChannel::name() { return "EVM"; }
 const char* ExecutiveWarnChannel::name() { return WarnChannel::name(); }
 
 StandardTrace::StandardTrace():
-	m_trace(make_shared<Json::Value>(Json::arrayValue))
+	m_trace(Json::arrayValue), m_levels(Json::arrayValue)
 {}
 
 bool changesMemory(Instruction _inst)
```
chriseth, line 43: please add delegatecall here

libethereum/Executive.cpp
```diff
@@ -66,9 +67,15 @@ void StandardTrace::operator()(uint64_t _steps, Instruction inst, bigint newMemS
```
chriseth, line 68: Can you change the `static_cast` above to a `dynamic_cast` on the reference?

libethereum/Executive.cpp
```diff
@@ -66,9 +67,15 @@ void StandardTrace::operator()(uint64_t _steps, Instruction inst, bigint newMemS
...
+	if (m_lastCallData != ext.data.toBytes())
+	{
+		m_lastCallData = ext.data.toBytes();
+		r["calldata"] = memDump(ext.data.toBytes(), 16);
+	}
+
 	Json::Value stack(Json::arrayValue);
 	for (auto const& i: vm.stack())
-		stack.append(toHex(toCompactBigEndian(i), 1));
+		stack.append(prettyU256(i));
```
chriseth: This seems to add some magic. We should only use the raw value here.

libethereum/Executive.cpp
```diff
@@ -66,9 +67,15 @@ void StandardTrace::operator()(uint64_t _steps, Instruction inst, bigint newMemS
 
 	Json::Value r(Json::objectValue);
 
+	if (m_lastCallData != ext.data.toBytes())
```
chriseth: This comparison is quite expensive. Perhaps it is better to check whether the call depth increased (see below)?

libethereum/Executive.cpp
```diff
@@ -81,13 +88,17 @@ void StandardTrace::operator()(uint64_t _steps, Instruction inst, bigint newMemS
 		assert(m_lastInst.size() == ext.depth);
 		m_lastInst.push_back(inst);
 		newContext = true;
+		m_levels.append(std::to_string(_steps));
+		r["levels"] = m_levels;
```
chriseth: What is the meaning of `levels`?

libethereum/Executive.cpp
```diff
@@ -114,7 +125,7 @@ void StandardTrace::operator()(uint64_t _steps, Instruction inst, bigint newMemS
 	{
 		Json::Value storage(Json::objectValue);
 		for (auto const& i: ext.state().storage(ext.myAddress))
-			storage[toHex(toCompactBigEndian(i.first), 1)] = toHex(toCompactBigEndian(i.second), 1);
+			storage[prettyU256(i.first)] = prettyU256(i.second);
```
chriseth: also `prettyU256`

libethereum/Executive.cpp
yann300: m_levels is an array which contain vmtrace index that correspond to depth change. This is used to build the callstack

libethereum/Executive.cpp
```diff
@@ -89,7 +84,8 @@ void StandardTrace::operator()(uint64_t _steps, Instruction inst, bigint newMemS
 		m_lastInst.push_back(inst);
 		newContext = true;
 		m_levels.append(std::to_string(_steps));
-		r["levels"] = m_levels;
+		r["levels"] = m_levels; // TODO: move the processing to JavaScript.
```
chriseth: Do you want to remove this line as part of this PR or as part of a future user story?

libethereum/Executive.cpp
```diff
@@ -89,7 +84,8 @@ void StandardTrace::operator()(uint64_t _steps, Instruction inst, bigint newMemS
...
+		r["calldata"] = memDump(ext.data.toBytes(), 16);
```
chriseth, line 86: Please don't use `memDump` (it performs some strange formatting), but rather use `toHex` as we use for the memory below.

PR #238, opened 2016-04-21
Assert that sealengine is there.

chriseth:


libethcore/SealEngine.cpp
```diff
@@ -51,6 +51,7 @@ void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, Transactio
 SealEngineFace* SealEngineRegistrar::create(ChainOperationParams const& _params)
 {
 	SealEngineFace* ret = create(_params.sealEngineName);
+	assert(!!ret && "Seal engine not found.");
```
bobsummerwill, line 54: !!ret ?

Why not just ret?

chriseth, line 54: yeah, might also work without `!!`. The `!!` is to force bool context, things like `assert(ret)` sometimes do not work on windows.

bobsummerwill, line 54: Well, the most correct form would actually be "(ret != NULL)"

chriseth, line 54: you mean `ret != nullptr` :-)

bobsummerwill, line 54: I do as of C++11, yes :-)

PR #241, opened 2016-04-23
Added null-pointer guard around setChainParams() calls

bobsummerwill:


libethereum/ChainParams.cpp
```diff
@@ -116,8 +116,9 @@ ChainParams ChainParams::loadGenesis(string const& _json, h256 const& _stateRoot
 SealEngineFace* ChainParams::createSealEngine()
 {
 	SealEngineFace* ret = SealEngineRegistrar::create(sealEngineName);
-	assert(ret && "Seal engine not found");
-	ret->setChainParams(*this);
+	assert((ret != nullptr) && "Seal engine not found");
+	if (ret != nullptr)
```
chriseth: The coding style actually recommends `!ret` here.

bobsummerwill: Why?   That is pretty poor advice, IMHO.
And where do you see that recommendation?

chriseth: It is not explicit in the CodingStandards.txt, but it is used like that all over the codebase.
Pro: Much better readable, goes well with type changes
Con: Goes well with type changes

libethereum/ChainParams.cpp
```diff
@@ -116,8 +116,9 @@ ChainParams ChainParams::loadGenesis(string const& _json, h256 const& _stateRoot
...
+		ret->setChainParams(*this);
 	if (sealRLP.empty())
 	{
 		sealFields = ret->sealFields();
```
chriseth, line 125: You use ret again here, so probably better to return early or throw.

libethcore/SealEngine.cpp
```diff
@@ -51,7 +51,8 @@ void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, Transactio
 SealEngineFace* SealEngineRegistrar::create(ChainOperationParams const& _params)
 {
 	SealEngineFace* ret = create(_params.sealEngineName);
-	assert(!!ret && "Seal engine not found.");
-	ret->setChainParams(_params);
+	assert(ret && "Seal engine not found.");
```
chfast, line 54: The assert is definitely a bad choice here if `create()` can fail.

libethcore/SealEngine.cpp
```diff
@@ -51,7 +51,8 @@ void SealEngineFace::verifyTransaction(ImportRequirements::value _ir, Transactio
...
+	if (ret)
```
chfast, line 55: I would prefer using "quick return" pattern as in other file.

libethereum/ChainParams.cpp
```diff
@@ -117,6 +117,8 @@ SealEngineFace* ChainParams::createSealEngine()
 {
 	SealEngineFace* ret = SealEngineRegistrar::create(sealEngineName);
 	assert(ret && "Seal engine not found");
+	if (!ret)
```
chfast, line 120: The same issue with assert.

libethereum/ChainParams.cpp
```diff
@@ -117,6 +117,8 @@ SealEngineFace* ChainParams::createSealEngine()
...
+		return NULL;
```
chfast: ~~NULL~~nullptr.

bobsummerwill:
RE:   Quick return.

Original code prior to @chriseth assertion:

```
SealEngineFace* ret = create(_params.sealEngineName);
ret->setChainParams(_params);
return ret;
```

Current code:

```
SealEngineFace* ret = create(_params.sealEngineName);
assert(ret && "Seal engine not found.");
ret->setChainParams(_params);
return ret;
```

My proposed code:

```
SealEngineFace* ret = create(_params.sealEngineName);
assert(ret && "Seal engine not found.");
if (ret)
    ret->setChainParams(_params);
return ret;
```

So what are you suggesting, @chfast?   This?   Is it any better?

```
SealEngineFace* ret = create(_params.sealEngineName);
assert(ret && "Seal engine not found.");
if (!ret)
    return NULL;
ret->setChainParams(_params);
return ret;
```

chfast:
In general I propose to remove the asserts. The rest is about coding style only.

bobsummerwill:
NOTE - So we had nothing.    Then we added an assert for better error messages in debug builds.   And now I am attempting to NOT crash.      If we really want to follow through here, we should add handling in the sites which call this too.    I've not looked at all those yet.    I'm just trying to improve the code we just modified, so we can get at least these functions in a state which we like.

bobsummerwill:
OK - over to you @chriseth :-)

chfast:
If asserts are there temporary for diagnostics then add a comment about that and leave them. And add an issue to track the bug later one.

chriseth:
So in order to fix this properly, we need to check the call sites.
I would also prefer to add an assert to guard against future sloppy usage.
This is not about finding the cause of a problem but rather about failing fast if the problem occurs or persists.

bobsummerwill:
Yes - having the assert AND the runtime check is not a bad thing.

Can we get this committed as-is, please, @chriseth?    It will make things better.

chfast:
Just replace `NULL` with `nullptr`.

bobsummerwill:
DONE

chfast:
So I'm ok with the change. Not sure if you need more confirmations.

PR #243, opened 2016-04-26
Update evmjit to consume recent CMake config changes

chfast:


bobsummerwill:
Is this in need of review and/or further iteration, @chfast?

LGTM.

chfast:
I wasn't able to get a clean build of that on Windows. But as it is merged we can think of that as done. In case of any issue spotted let me know.

PR #246, opened 2016-04-27
Reformat code with clang-format.

chfast:
This is example of using clang-format. Config in .clang-format file. I used version 3.6. 3.8 seems to have more options: http://llvm.org/releases/3.8.0/tools/clang/docs/ClangFormatStyleOptions.html


bobsummerwill:
Great - thanks for sharing, @chfast :-)

And what is your command-line for the invocation here?   On an Ubuntu box, I assume?

Do you know if `clang-format` is supported in AppleClang, or would we need to rely on a brew install of "real clang"?    What is happening with Clang for Windows these days?   Thanks!

gcolvin:
Great, thanks.  Seems best not to run it against existing code, lest we drown in diffs.  But for new code it will help me.  Is there an option to use spaces for interior fill?  Otherwise, Client.h looks like this in my editor with 4-space tabs

```
    bool remoteActive() const;                            ///< Is there an active and valid remote worker?
    bool m_remoteWorking = false;                         ///< Has the remote worker recently been reset?
    std::atomic<bool> m_needStateReset = {false};         ///< Need reset working state to premin on next sync
    std::chrono::system_clock::time_point m_lastGetWork;  ///< Is there an active and valid remote worker?
```

but looks like this here on GitHub (and with most any of my command-line tools) using 8-space tabs

```
        bool remoteActive() const;                                                        ///< Is there an active and valid remote worker?
        bool m_remoteWorking = false;                                             ///< Has the remote worker recently been reset?
        std::atomic<bool> m_needStateReset = {false};             ///< Need reset working state to premin on next sync
        std::chrono::system_clock::time_point m_lastGetWork;  ///< Is there an active and valid remote worker?
```

chfast:
1. clang-format is standalone executable, but in case of Ubuntu it might be a part of clang package.
2. You can find integration for almost every editor in the world, even Visual Studio. It is also easy to add integration manually by assigning keyboard shortcut.
3. Yes, we should not modify existing code.
4. The configuration needs more tweaks. I didn't have time to check all options.
5. @gcolvin You can force GitHub to display tab of size 4 by appending `?ts=4` to an URL.

bobsummerwill:
Hey, hey!   Check it out on OS X:

```
brew install clang-format
```

chriseth:
@chfast can this be closed?

chfast:
Give me a bit more time on that.

bobsummerwill:
Yes, let's not close this.

Having an optional formatting tool would be useful, even if it's a "slow burn" activity.

PR #254, opened 2016-05-11
Multiple BALANCE instruction call test

winsvega:


chriseth:
What is the purpose of this test? Why did the test run fail?

winsvega:
test run fails due to some other reason (Jenkins)
Paweł Bylica @chfast  asked for this tests
never have enough of them :)

chriseth:
@winsvega please still explain what it does.

winsvega:
Pawel wanted a test in which BALANCE instruction is being called at least twice during one call.

chfast:
I needed such test.

chfast:
@chriseth any objections? can we merge it?

PR #255, opened 2016-05-17
fix #515: error: ignoring attributes on template argument 'cl_int {aka int}' 

moneroexamples:
On Arch with gcc 6.1 compilation of ethereum fails do to:

```
/webthree-umbrella/libethereum/libethash-cl/cl.hpp:2423:28: error: ignoring attributes on template argument 'cl_int {aka int}' [-Werror=ignored-attributes]
         VECTOR_CLASS<cl_int>* binaryStatus = NULL,
```

https://github.com/ethereum/webthree-umbrella/issues/515

It can be fixed by adding new pragam to cl.hpp

```
#pragma GCC diagnostic ignored "-Wignored-attributes"
```

### DIFF
--- a/libethash-cl/cl.hpp
+++ b/libethash-cl/cl.hpp
@@ -22,7 +22,8 @@
  ******************************************************************************/
 
 #pragma GCC diagnostic ignored "-Wdeprecated-declarations"
-
+#pragma GCC diagnostic ignored "-Wignored-attributes"
+ 
 /*! \file
  *
  *   \brief C++ bindings for OpenCL 1.0 (rev 48) and OpenCL 1.1 (rev 33)    


bobsummerwill:
Thanks, @moneroexamples!

See http://52.28.164.97/job/project-build/label=master/4821/.

Looks like the new pragma will need wrapping in some GCC version conditionals, eh?   To avoid breaking earlier GCC versions, like the one we have on our Ubuntu builds.

So I guess it's going to need nested ... "if GCC" and then "if GCC > 6.1.1" or similar?

5chdn:
Don't want to create a new PR, try:

```
#if __GNUC__ >= 6 
    #pragma GCC diagnostic ignored "-Wignored-attributes" 
#endif
```

...but this would only work on *nix systems I guess.

moneroexamples:
With the `__GNUC__ >= 6` condition proposed by  @5chdn the compilation works fine on Arch and GCC 6.1. 

Should I make a new commit with this change?

5chdn:
Yes please, let's see if the build server likes this, too :p

moneroexamples:
I close this pull request, as new one with the proposed `__GNUC__ >= 6` condition was created:
https://github.com/ethereum/libethereum/pull/256

PR #257, opened 2016-05-19
Reject blocks above max block gas.

chriseth:
This might need some test modifications.


winsvega:
in Greg's PR would be a conflict

chriseth:
@winsvega please explain a bit more.

winsvega:
nevermind. merge please

PR #258, opened 2016-05-19
test client

winsvega:
work in progress for TestClient implementation
connects to https://github.com/ethereum/webthree-umbrella/issues/260


libethereum/ClientTest.cpp
```diff
@@ -0,0 +1,155 @@
...
+			UpgradeGuard l2(l);
+			if (!m_working.sealBlock(s.out()))
+				return false;
+		}
+		DEV_WRITE_GUARDED(x_postSeal)
+			m_postSeal = m_working;
+		newBlock = m_working.blockData();
+	}
+
+	// OPTIMISE: very inefficient to not utilise the existing OverlayDB in m_postSeal that contains all trie changes.
```
chriseth, line 101: Is there a way to do this without having to copy most of the code (i.e. split the function you copied from into more than one or add a parameter)?

winsvega, line 101: this would require to change Client.h
for instance I could move the code into protected private method and just execute a call to it from onNewBlocks.

like:

```
virtual onNewBlocks()
{
     /* any additional code here */
     internalOnNewBlocks();
}
```

chriseth, line 101: Yes, please change client.h but please also group code into meaningful chunks, i.e. not just call the function `internalOnNewBlocks` if it can be avoided.
Oh and you can also directly call functions of parent classes, right?

```
virtual onNewBlocks()
{
    Client::onNewBlocks();
    /* now do the new stuff */
}
```

libethereum/ClientTest.cpp
```diff
@@ -0,0 +1,155 @@
...
+	Client(_params, _networkID, _host, _gpForAdoption, _dbPath, _forceAction, _limits)
+{
+}
+
+bool ClientTest::addBlock(string const& _rlp)
+{
+	if (auto h = m_host.lock())
+		h->noteNewBlocks();
+
+	bytes rlpBytes = fromHex(_rlp.substr(0, 2) == "0x" ? _rlp.substr(2) : _rlp, WhenError::Throw);
```
chriseth, line 59: I think the check for `"0x"` is not needed, that should be done inside `fromHex`

libethereum/ClientTest.cpp
```diff
@@ -0,0 +1,155 @@
...
+	m_blockMineTimestamp = _timestamp;
+	startSealing();
+}
+
+bool ClientTest::submitSealed(bytes const& _header)
+{
+	BlockHeader previousHeader(bc().block());
+	BlockHeader header(_header, HeaderData);
+
+	header.setTimestamp(previousHeader.timestamp() + m_blockMineTimestamp);
```
chriseth, line 81: After we modify the timestamp in the header, all transactions need to be replayed (because they can access the timestamp and have a different result dependeing on that). Where is that done?

PR #259, opened 2016-05-22
Remove Frontier canaries, which were left in by mistake.

bobsummerwill:


chriseth:
@bobsummerwill please also remove the "sealOnBadChain" commandline option in eth.

bobsummerwill:
Sure, will do, @chriseth.

Do you see any other "old fluff" like this which we should probably eliminate?   I'm looking at Olympic support, for example.

chriseth:
Yes, we can remove olympic from the commandline options, but please keep the parameters in the codebase.

PR #265, opened 2016-05-26
Functionality for tests via rpc.

chriseth:
replaces #258

Connects to https://github.com/ethereum/webthree-umbrella/issues/260


bobsummerwill:
These "breaks" are because of our broken cross-repo Jenkins support, right?

chriseth:
@bobsummerwill no, this PR should build fine, as it is at the "deepest level".

PR #268, opened 2016-05-31
Provide access to secret store without cache.

chriseth:


chfast:
This looks ok, however the interface is getting uglier :)

Please check CI failures btw.

chriseth:
I cannot imagine how the test failures could be related to the secret store.

PR #234, opened 2016-04-15
Make the interpreter scream...

gcolvin:
- Eliminate unnecessary bigint.
- Eliminate gas and memory counting overhead for operations that don't need it.
- Parameterize size of machine word. Slightly faster with 64-bit word, but fails more tests of extreme parameters.
- More improvements to come.

Have achieved about a 3X speedup so far. I see two big impediments to the next round of improvement.

First, the available backends for boost::multiprecision are optimized for very large numbers. So for our 256-bit numbers we use 320 bits to allow for a header. That means our numbers won't fit evenly in cache lines and memory pages. Plus boost does not support exchange with external binary formats. So we need to make a fixed-size replacement that uses only 256- (64-, 128-, or 512-) bits, and has fast support for external formats.
Second, the JUMP instruction is a computed goto: it takes its argument off of the stack. Usually the destination is fixed, pushed (slowly, see above) from the code preceding the jump. But because the destination might not be fixed it must be validated against a table of valid destinations. The overhead is bad enough that in the following loop about half the time is spent in the loop execution itself.
    for (int i = 0; i < 100000; ++i) {
       rand1 = 0xffe7649d5eca8417 \* rand1 + 0xf47ed85c4b9a6379;
       rand2 = 0xf8e5dd9a5c994bba \* rand2 + 0xf91d87e4b8b74e55;
       rand3 = 0xff97f6f3b29cda52 \* rand3 + 0xf393ada8dd75c938;
       rand4 = 0xfe8d437c45bb3735 \* rand3 + 0xf47d9a7b5428ffec;
    }
Better wide integers requires coding the usual tricky algorithms for arithmetic operations, plus all the other functions required of a boost::multiprecision backend. Then it can be plugged into the classes in libdevcore/common.h to speed up our extra-wide integers in the interpreter and elsewhere.
Dealing with computed JUMP addresses without a table search is harder. Faster loading of the jump adress from the bytecode array will help, but a search for a valid destination seems unavoidable.

PR #270, opened 2016-05-31
Make the interpreter whimper

gcolvin:
This is the continuation of https://github.com/ethereum/libethereum/pull/234, and one phase in the ongoing project of improving the performance of the EVM interpreter.


chriseth:
This pull request seems to already include your previous changes. I think you have to rebase it on top of latest develop.

gcolvin:
I have now pushed a rebased and deconflicted libethereum.  Of course it cannot build until the small mix and alethzero patches in ethereum/mix#310 and ethereum/alethzero#165 are also in.

chriseth:
Thanks, will take a look next week!

gcolvin:
Notes for reviewers.

The first major change to the interpreter was to reduce the size of arithmetic types.  The io_gas reference, stack slots, and more are 256-bit unsigned integers.  Calculations throughout were carried out with at least 256-bit integers, and in cases where overflow was possible with arbitrary-precision bigint.

In the real world computers have 32 and 64-bit registers, so emulating 256-bit and wider computations has its costs.  And in the real world 64-bits worth is more gas than will ever be affordable.  So much of the gas calculation has been changed to use 64-bit values.  For the necessarily 256-bit computations of the VM itself it is convenient to use bigint as the overflow type, but inefficient.  In most cases 512-bits suffices, and bigint is now used only in the calculation of exponentials.

The next major change was to the staging of gas, stack, and memory computations.  The original interpreter had a checkRequirements() function that was called for each opcode in the main interpreter loop to check for gas, memory, and stack usage, and to call the tracing plugin.  This made it easier to verify that these calculations conformed to the complex rules in the Yellow Paper.  Unfortunately, in most cases the checkRequirements() routine was more expensive to run than the code to compute the opcode itself.  So checkRequirements() is gone, factored into a small collection of closures in execImpl().  These are called only as relevant to each opcode.

The most important minor change was to switch from handling the stack via std::vector functions, (which could incur a check for overflow on each operation, and indirection to the vector as a member variable) to pulling the raw stack pointer out of the std::vector, replacing member functions with raw arithmetic on the stack pointer. 

On two small benchmarks I've been using to maintain forward progress - a random number generator with four MULs and four ADDs per loop, and an empty loop - we see speedups of 6X and 9X respectively.

This speed was obtained at some cost in stack usage, so a round of work ensued to get less commonly used functions that used stack into VMNoInline.cpp and to generally organize execImpl() so that the compiler would not become overwhelmed and stop allocating temporaries intelligently.  At this point the new interpreter uses a bit less stack than the original.

There remains plenty of room for improvement. The recent decision to restrict gas limits to 2^63-1 should create more opportunity to avoid overflow.  A trampoline design for opcodes that call out could further reduce stack usage.  Precomputing valid jump destinations could eliminate the overhead of verifying them for every jump.  Most promising, dropping boost::multi precision in favor of the assembly-level Gnu multi-precision routines should give a further increase in efficiency, if only by narrowing stack slots from 5 words to 4.

CMakeLists.txt
```diff
@@ -8,7 +8,7 @@ include(EthPolicy)
 eth_policy()
 
 # project name and version should be set after cmake_policy CMP0048
-project(ethereum VERSION "1.2.6")
+project(ethereum VERSION "1.2.4")
```
chriseth, line 11: Rebase artifact.

chriseth:
You changed the file permissions to "executable" on several files.

libethereum/Client.h
```diff
@@ -330,4 +330,4 @@ class Client: public ClientBase, protected Worker
 };
 
 }
-}
```
chriseth, line 333: Please take the file out of the changeset completely.

libethereum/Executive.cpp
```diff
@@ -127,14 +126,10 @@ void StandardTrace::operator()(uint64_t _steps, Instruction inst, bigint newMemS
 		r["storage"] = storage;
 	}
 
-	if (returned || newContext)
```
chriseth, line 130: We made several changes in these files during the last weeks - did you actually intend to do these changes apart from the PC change?

gcolvin:
I've no internet until ISP fixes it, so came to town to use library briefly.  I don't know how I would have made anything executable.  Certainly never ran chmod.  But rebase threw away everything Dmitry and I ever did to my repo, so it was hand copying and git foo in the end.  I expect it's a mess.

libethereum/Executive.cpp
gcolvin, line 130: I only intended the changes for the PC arg.

chriseth:
Replaced by #277

PR #271, opened 2016-06-01
testClient reset chainParams

winsvega:
connects to https://github.com/ethereum/webthree-umbrella/issues/542

### DIFF
--- a/libethereum/ClientTest.cpp
+++ b/libethereum/ClientTest.cpp
@@ -56,14 +56,14 @@ void ClientTest::setChainParams(string const& _genesis)
 	{
 		params = params.loadConfig(_genesis);
 		if (params.sealEngineName != "NoProof")
-			BOOST_THROW_EXCEPTION(ChainParamsNotNoProof() << errinfo_comment("provided configuration is not well formatted"));
+			BOOST_THROW_EXCEPTION(ChainParamsNotNoProof() << errinfo_comment("Provided configuration is not well formatted."));
 
 		reopenChain(params, WithExisting::Kill);
 		setAuthor(params.author); //for some reason author is not being set
 	}
 	catch (...)
 	{
-		BOOST_THROW_EXCEPTION(ChainParamsInvalid() << errinfo_comment("provided configuration is not well formatted"));
+		BOOST_THROW_EXCEPTION(ChainParamsInvalid() << errinfo_comment("Provided configuration is not well formatted."));
 	}
 }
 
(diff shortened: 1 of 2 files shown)


libethereum/Client.h
```diff
@@ -286,7 +286,7 @@ class Client: public ClientBase, protected Worker
 	/// @warning May be called from any thread.
 	void onBadBlock(Exception& _ex) const;
 
-	BlockChain m_bc;						///< Maintains block database and owns the seal engine.
+	std::unique_ptr<BlockChain> m_bc;		///< Maintains block database and owns the seal engine.
```
chriseth: I think we cannot make that a pointer - I would think that too many objects hold references to structures inside this.

In general, closing and reopening the database are really expensive operations we should avoid. Perhaps it is easier to just modify the genesis block (i.e. not all the parameters)?

I think it is fine to just add a function to BlockChain that changes `m_params` and resets some things.

libethereum/Client.h
```diff
@@ -204,8 +204,8 @@ class Client: public ClientBase, protected Worker
 	void init(p2p::Host* _extNet, std::string const& _dbPath, WithExisting _forceAction, u256 _networkId);
 
 	/// InterfaceStub methods
-	BlockChain& bc() override { return m_bc; }
-	BlockChain const& bc() const override { return m_bc; }
+	BlockChain& bc() override { return const_cast<BlockChain&>(*m_bc.get()); }
```
chriseth: Why do you need the `const_cast`?

libethereum/ClientTest.cpp
```diff
@@ -63,7 +82,7 @@ bool ClientTest::addBlock(string const& _rlp)
 
 void ClientTest::rewindToBlock(unsigned _number)
 {
-	m_bc.rewind(_number);
+	m_bc.get()->rewind(_number);
```
chriseth: No need to call `get()` here.

libethereum/ClientTest.cpp
```diff
@@ -48,7 +47,24 @@ ClientTest::ClientTest(
...
+		params = params.loadConfig(_genesis);
+		if (params.sealEngineName != "NoProof")
+			BOOST_THROW_EXCEPTION(ChainParamsNotNoProof() << errinfo_comment("provided configuration is not well formatted"));
+
+		reopenChain(params, WithExisting::Kill);
+		setAuthor(params.author); //for some reason author is not being set
+	}
+	catch (...)
+	{
+		BOOST_THROW_EXCEPTION(ChainParamsInvalid() << errinfo_comment("provided configuration is not well formatted"));
```
chriseth: Please use a full sentence, i.e. start with `P` and end in `.` (also above).

chriseth:
Looks good apart from the minor comment.

PR #272, opened 2016-06-02
Do not provide Keccak function to EVMJIT

chfast:
EVMJIT has added its own Keccak implementation. Experting one form libethereum is not needed any more.


chfast:
Please allow authors to merge theirs PRs and take responsibility for them. This one will fail building on Windows. The fix is waiting for Jenkins.

PR #273, opened 2016-06-03
Remove unused variables

enriquefynn:
Failed when compiling with -Wfor-loop-analysis


bobsummerwill:
Thanks!   Merged.    Will just update the umbrella with this change too.

bobsummerwill:
See http://52.28.164.97/job/update-umbrella/.    They're just running now.

PR #277, opened 2016-06-08
Interpreter speedup

chriseth:
This fixes rebase artifacts from #270 . This still needs to be reviewed (also by me). Replaces #270.


chriseth:
Copying comment by @gcolvin:

Notes for reviewers.

The first major change to the interpreter was to reduce the size of arithmetic types.  The io_gas reference, stack slots, and more are 256-bit unsigned integers.  Calculations throughout were carried out with at least 256-bit integers, and in cases where overflow was possible with arbitrary-precision bigint.

In the real world computers have 32 and 64-bit registers, so emulating 256-bit and wider computations has its costs.  And in the real world 64-bits worth is more gas than will ever be affordable.  So much of the gas calculation has been changed to use 64-bit values.  For the necessarily 256-bit computations of the VM itself it is convenient to use bigint as the overflow type, but inefficient.  In most cases 512-bits suffices, and bigint is now used only in the calculation of exponentials.

The next major change was to the staging of gas, stack, and memory computations.  The original interpreter had a checkRequirements() function that was called for each opcode in the main interpreter loop to check for gas, memory, and stack usage, and to call the tracing plugin.  This made it easier to verify that these calculations conformed to the complex rules in the [Yellow Paper](https://github.com/ethereum/yellowpaper).  Unfortunately, in most cases the checkRequirements() routine was more expensive to run than the code to compute the opcode itself.  So checkRequirements() is gone, factored into a small collection of closures in execImpl().  These are called only as relevant to each opcode.

The most important minor change was to switch from handling the stack via std::vector functions, (which could incur a check for overflow on each operation, and indirection to the vector as a member variable) to pulling the raw stack pointer out of the std::vector, replacing member functions with raw arithmetic on the stack pointer. 

On two small benchmarks I've been using to maintain forward progress - a random number generator with four MULs and four ADDs per loop, and an empty loop - we see speedups of 6X and 9X respectively.

This speed was obtained at some cost in stack usage, so a round of work ensued to get less commonly used functions that used stack into VMNoInline.cpp and to generally organize execImpl() so that the compiler would not become overwhelmed and stop allocating temporaries intelligently.  At this point the new interpreter uses a bit less stack than the original.

There remains plenty of room for improvement. The recent decision to restrict gas limits to 2^63-1 should create more opportunity to avoid overflow.  A trampoline design for opcodes that call out could further reduce stack usage.  Precomputing valid jump destinations could eliminate the overhead of verifying them for every jump.  Most promising, dropping boost::multi precision in favor of the assembly-level Gnu multi-precision routines should give a further increase in efficiency, if only by narrowing stack slots from 5 words to 4.

gcolvin:
The diffs are hard to follow with so many changes, but the bulk of the work was moving the main interpreter class from the original version

> https://github.com/ethereum/libethereum/blob/develop/libevm/VM.h
> https://github.com/ethereum/libethereum/blob/develop/libevm/VM.cpp

to the changed version described above

> https://github.com/chriseth/libethereum/blob/develop/libevm/VM.h
> https://github.com/chriseth/libethereum/blob/develop/libevm/VM.cpp

gcolvin:
Note that this PR cannot build without the small mix and alethzero patches in ethereum/mix#310 and ethereum/alethzero#165.

gcolvin:
@ chriseth  This commit looks good.  Much clearer names.  I am confused by the diffs from rmword, as there were no uses of rmword or the rest remaining in #270.

gcolvin:
As for

```
            // TODO Is that needed?
            memMax(*SP);
            // TODO: overflow?
            // TODO (similar cases below)
            newMemorySize = toUint64(*SP) + 32;
```

I think that memMax() is not needed.  Overflow in these cases could be handled instead by doing the addition to the u256 before truncating to uint64_t.  So it could it just be:

```
            newMemorySize = toUint64(*SP + 32);
```

But given the memMax() call there is no chance of overflowing 64 bits in the addition, and the gas will be checked later, so it could also be

```
            memMax(*SP);
            newMemorySize = uint64(*SP) + 32;
```

I'm not sure which is faster.

chriseth:
Oh, I see! We can have an overflow of 63 bits, but not of 64 bits. Ok!

libevm/VM.cpp
```diff
@@ -76,181 +76,193 @@ template <class S> S modWorkaround(S const& _a, S const& _b)
...
-	m_onFail = std::function<void()>(onOperation);
-	
-	auto checkStack = [&](unsigned _n, unsigned _d)
-	{
-		const int size = 1 + SP - m_stack, usedSize = size - _n;
-		if (usedSize < 0 || usedSize + _d > 1024)
-			throwVMStackException(size, _n, _d);
-	};
+	// trampoline to minimize depth of call stack when calling out
+	m_bounce = &VM::interpretCases;
```
chriseth, line 154: Does this have to be a member? What about the following:

```
auto continuation = &VM::interpretCases;
do
  continuation = (this->*continuation)();
while (continuation);
```

winsvega:
running tests locally on ubuntu. results are fine

PR #278, opened 2016-06-09
ethvm use input for transaction and code for contract

winsvega:
connects to https://github.com/ethereum/webthree-umbrella/issues/555

### DIFF
--- a/ethvm/main.cpp
+++ b/ethvm/main.cpp
@@ -38,17 +38,21 @@ using namespace std;
 using namespace dev;
 using namespace eth;
 
+static const u256 MaxBlockGasLimit = ChainParams(genesisInfo(Network::HomesteadTest)).u256Param("maxGasLimit");
+
 void help()
 {
 	cout
 		<< "Usage ethvm <options> [trace|stats|output] (<file>|--)" << endl
 		<< "Transaction options:" << endl
 		<< "    --value <n>  Transaction should transfer the <n> wei (default: 0)." << endl
-		<< "    --gas <n>  Transaction should be given <n> gas (default: block gas limit)." << endl
-		<< "    --gas-limit <n>  Block gas limit (default: " << DefaultBlockGasLimit << "." << endl
+		<< "    --gas <n>    Transaction should be given <n> gas (default: block gas limit)." << endl
+		<< "    --gas-limit <n>  Block gas limit (default: " << MaxBlockGasLimit << ")." << endl
 		<< "    --gas-price <n>  Transaction's gas price' should be <n> (default: 0)." << endl
 		<< "    --sender <a>  Transaction sender should be <a> (default: 0000...0069)." << endl
 		<< "    --origin <a>  Transaction origin should be <a> (default: 0000...0069)." << endl
+		<< "    --input <d>   Transaction code should be <d>" << endl
+		<< "    --code <d>    Contract code <d>. Makes transaction a call to this contract" << endl
 #if ETH_EVMJIT
 		<< endl
 		<< "VM options:" << endl
@@ -84,25 +88,24 @@ enum class Mode
 
 int main(int argc, char** argv)
 {
-	string incoming = "--";
-
 	Mode mode = Mode::Statistics;
 	VMKind vmKind = VMKind::Interpreter;
 	State state(0);
 	Address sender = Address(69);
 	Address origin = Address(69);
 	u256 value = 0;
-	u256 gas = DefaultBlockGasLimit;
+	u256 gas = MaxBlockGasLimit;
 	u256 gasPrice = 0;
 	bool styledJson = true;
 	StandardTrace st;
 	EnvInfo envInfo;
 	Network networkName = Network::HomesteadTest;
-	envInfo.setGasLimit(gas);
+	envInfo.setGasLimit(MaxBlockGasLimit);
+	string transactionDataString = "";
+	string contractCodeString = "";
 	
 	Ethash::init();
 	NoProof::init();
-
 
 	for (int i = 1; i < argc; ++i)
 	{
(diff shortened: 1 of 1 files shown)


chfast:
It works (however you have wrong call/create selection condition).

ethvm/main.cpp
```diff
@@ -84,26 +88,25 @@ enum class Mode
...
-	u256 gas = DefaultBlockGasLimit;
+	u256 gas = MaxBlockGasLimit;
 	u256 gasPrice = 0;
 	bool styledJson = true;
 	StandardTrace st;
 	EnvInfo envInfo;
 	Network networkName = Network::HomesteadTest;
-	envInfo.setGasLimit(gas);
+	envInfo.setGasLimit(MaxBlockGasLimit);
+	string transactionDataString = "";
```
chfast: std::string does not need initializer.

ethvm/main.cpp
```diff
@@ -181,29 +178,35 @@ int main(int argc, char** argv)
...
-	bytes code;
-	if (incoming == "--" || incoming.empty())
-		for (int i = cin.get(); i != -1; i = cin.get())
-			code.push_back((char)i);
-	else
-		code = contents(incoming);
-	bytes data = fromHex(boost::trim_copy(asString(code)));
-	if (data.empty())
-		data = code;
+	bytes contractCode;
```
chfast: Unused

ethvm/main.cpp
```diff
@@ -181,29 +178,35 @@ int main(int argc, char** argv)
...
+	bytes transactionDataCode = fromHex(transactionDataString);
+	Transaction t = eth::Transaction(value, gasPrice, gas, transactionDataCode, 0);
+	Address contractDestination("1122334455667788991011121314151617181920");
+	if (contractCodeString != "")
+	{
+		contractCode = fromHex(contractCodeString);
```
chfast: Unused.

ethvm/main.cpp
```diff
@@ -223,7 +226,11 @@ int main(int argc, char** argv)
 	};
 
 	executive.initialize(t);
-	executive.create(sender, value, gasPrice, gas, &data, origin);
+	if (contractCodeString != "")
```
chfast: Should be `==`. Preferably `.empty()`.

ethvm/main.cpp
```diff
@@ -181,29 +178,35 @@ int main(int argc, char** argv)
...
-	else
-		code = contents(incoming);
-	bytes data = fromHex(boost::trim_copy(asString(code)));
-	if (data.empty())
-		data = code;
+	bytes contractCode;
+	bytes transactionDataCode = fromHex(transactionDataString);
+	Transaction t = eth::Transaction(value, gasPrice, gas, transactionDataCode, 0);
+	Address contractDestination("1122334455667788991011121314151617181920");
+	if (contractCodeString != "")
```
chfast, line 214: Preferably `!.empty()`.

ethvm/main.cpp
```diff
@@ -181,29 +178,35 @@ int main(int argc, char** argv)
...
+		else if (arg == "--input" && i + 1 < argc)
+			transactionDataString = argv[++i];
+		else if (arg == "--code" && i + 1 < argc)
+			contractCodeString = argv[++i];
 	}
 
 	VMFactory::setKind(vmKind);
 
-	bytes code;
-	if (incoming == "--" || incoming.empty())
```
chfast, line 191: I'm not sure we can remove that. I think @gcolvin is using this feature.

winsvega, line 191: so if --call and --input are not set we assume that input data for transaction is passed either by filepath or cin input?

chfast, line 191: `--input` and filepath arguments are independent, see https://github.com/ethereum/libethereum/pull/278/files#diff-95f10d2c41ff73513a9d323e1d0fa531L184. We just need to decide what to do if both are present. I suggest print an error end exit.

ethvm/main.cpp
```diff
@@ -226,7 +246,7 @@ int main(int argc, char** argv)
 	};
 
 	executive.initialize(t);
-	if (contractCodeString != "")
+	if (!contractCodeString.empty())
```
chfast, line 249: This condition is still the opposite. We should go create mode if the code string **is empty**.

chfast:
One, but important character is wrong! Besides that looks good.

chfast:
Do you know why the tests fail? Is it related?

ethvm/main.cpp
```diff
@@ -182,17 +183,36 @@ int main(int argc, char** argv)
...
 	}
 
 	VMFactory::setKind(vmKind);
 
-	bytes contractCode;
-	bytes transactionDataCode = fromHex(transactionDataString);
+
+	//Read transaction data field from input
+	bytes code;
+	if (incoming == "--" || incoming.empty())
```
chfast, line 195: I missed that. It should be only `if (incoming == "--")`. In case `incoming` is empty we should assume no input is provided by standard input. Otherwise the program waits for input in cases it is not expected.

ethvm/main.cpp
winsvega, line 249: no. code is for contract code.  input is for transaction data

chfast, line 249: That makes absolutely no sense.
- `--code A --input B` will create a contract with init code A, input B is ignored,
- `--input B` will call empty code with input B.
- `--code A` will create a contract with init code A.

winsvega:
http://52.28.164.97/job/project-test/label=master/2739/

can't see test results. must be some jenkins issue

ethvm/main.cpp
winsvega, line 249: wait. by  "will create a contract with init code A" you mean it will execute transaction that will create contract by code A?  

OR 

actual creation of a contract by using setCode  directly in the state?

PR #279, opened 2016-06-09
Update evmjit sub-module to include "CMake: set policy CMP0054 only if known" fix which gets Debian working.

bobsummerwill:
This bug fix gets Debian working.


chfast:
Merge if green.

One suggestion. Can you make commit messages in the title / details fusion? Title should be no longer than ~60 chars, no full stop.

bobsummerwill:
Thanks, @chriseth!

PR #280, opened 2016-06-10
Revert "Remove unused variables"

chriseth:
Reverts ethereum/libethereum#273

The loop variable is actually modified inside the callback.

### DIFF
--- a/ethminer/MinerAux.h
+++ b/ethminer/MinerAux.h
@@ -523,7 +523,7 @@ private:
 					solution = sol;
 					return completed = true;
 				});
-				while(true)
+				for (unsigned i = 0; !completed; ++i)
 				{
 					auto mp = f.miningProgress();
 					f.resetMiningProgress();


bobsummerwill:
I have a better fix suggestion to make.   Give me 2 mins and I will make a new PR.

CC @enriquefynn, @sammy007, @awrelll

Aside - the change was an attempted fix for warnings which got the client working on FreeBSD (but obviously with this horrible knock-on because of uncertain side-effects.

We should probably make a 1.2.8 release today, @chriseth, both for this fix, and for the Mix Debug OS X.

bobsummerwill:
See https://github.com/ethereum/libethereum/pull/281.    I would suggest that we go with that alternative fix, @chriseth.

bobsummerwill:
I've tagged this as "got issues" because while it would fix mining, it would also break the newly working FreeBSD build.

#281 would, I believe, fix mining while not breaking FreeBSD.

chriseth:
Closing in favour of #281

PR #289, opened 2016-06-16
Switching some exception classes to use boost versions to resolve ARM Linux build breaks.

bobsummerwill:
CC @anthony-cros.

The classes in question are **exception_ptr**, **current_exception** and **rethrow_exception**, all of which are ONLY used in a single function within ExtVM.cpp.   Using the boost versions, rather than the std versions, allows us to work around a long-standing GLIBC issue where these classes are unavailable for architectures with incomplete atomic int support.

"std::exception_ptr is missing on architectures with incomplete atomic int support"
https://gcc.gnu.org/bugzilla/show_bug.cgi?id=58938

"std::exception_ptr not available for old x86 arch"
http://trac.wxwidgets.org/ticket/16634

"NDK: unable to use std::exception_ptr with GCC 4.8 on armeabi, GNU STL"
https://code.google.com/p/android/issues/detail?id=62648

### DIFF
--- a/libethereum/ExtVM.cpp
+++ b/libethereum/ExtVM.cpp
@@ -27,7 +27,7 @@
 using namespace dev;
 using namespace dev::eth;
 
-namespace
+namespace // anonymous
 {
 
 static unsigned const c_depthLimit = 1024;
@@ -59,7 +59,7 @@ void goOnOffloadedStack(Executive& _e, OnOpFunc const& _onOp)
 
 	// Create new thread with big stack and join immediately.
 	// TODO: It is possible to switch the implementation to Boost.Context or similar when the API is stable.
-	std::exception_ptr exception;
+	boost::exception_ptr exception;
 	boost::thread{attrs, [&]{
 		try
 		{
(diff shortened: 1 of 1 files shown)


chriseth:
Thanks, looks good!

chriseth:
Please rebase.

bobsummerwill:
Sorry - missed your "please rebase" comment and the fact that I had NOT done so prior to merging.   Apologies for noise in the history.

anthony-cros:
Wow @bobsummerwill , can't believe that hack had such a long life! It's a very good sign that you are now in a place to address the underlying problem, it seemed pretty far off at the time :)

bobsummerwill:
Ha ha!

So I am on vacation for the next 3 weeks or so, and I plan to try to blast my way through ALL of these:

https://github.com/doublethinkco/cpp-ethereum-cross/issues/11

Yes, that issue is still open and most of them are still in place :-)

However, I have a huge amount more context of the codebase now (as well as the ability to work through the upstreaming and review process quicker, because I can sit on both sides of the process).

We are getting very close to doing the repo-reorg to get back into cpp-ethereum, and I want to see if I can resolve these hacks on that same kind of timeline (or close to it), so that cpp-ethereum can have both native and cross-builds, and most or all of the technical debt to support that is gone).

I just got TESTS=On and TOOLS=On 95% of the way to done.    Just need to cross-build `libminiupnc` because it is used in `ethminer` and then all the executables will cross-build.

Guess what else is still an open issue?

https://github.com/doublethinkco/cpp-ethereum-cross/issues/20

Yes, fucking Tizen.     I still don't have the sodding code working on my Gear S2 :-)

I've been 100% on Foundation stuff since joining in February.     Time to pick this stuff back up too.

Best wishes!

PR #291, opened 2016-06-16
Weakening off an assert to a log warning in BlockChainSync

bobsummerwill:
That assertion is causing frequent crashes on Ubuntu.
Block headers map is empty, but block bodies map is not.  Force-clearing.
Fixes ethereum/webthree-umbrella#581

### DIFF
--- a/libethereum/BlockChainSync.cpp
+++ b/libethereum/BlockChainSync.cpp
@@ -672,7 +672,11 @@ void BlockChainSync::onPeerNewBlock(std::shared_ptr<EthereumPeer> _peer, RLP con
 		removeItem(m_bodies, blockNumber);
 		if (m_headers.empty())
 		{
-			assert(m_bodies.empty());
+			if (!m_bodies.empty())
+			{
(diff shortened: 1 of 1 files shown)


libethereum/BlockChainSync.cpp
```diff
@@ -672,7 +672,11 @@ void BlockChainSync::onPeerNewBlock(std::shared_ptr<EthereumPeer> _peer, RLP con
 		removeItem(m_bodies, blockNumber);
 		if (m_headers.empty())
 		{
-			assert(m_bodies.empty());
+			if (!m_bodies.empty())
+			{
+				clog(NetMessageDetail) << "Block headers map is empty, but block bodies map is not. Force-clearing."
```
chriseth: `;` at the end.

chriseth:
Looks good apart from the syntax error.

libethereum/BlockChainSync.cpp
bobsummerwill: Ugh.   Fixed.   Thanks.

PR #292, opened 2016-06-16
Take transactions from postSeal.

chriseth:


winsvega:
have you tested it with pending transactions?

chriseth:
No, but it works fine with the solidity tests, because we do not yet use that feature. Perhaps you could write some tests in javascript using ethereum-console?

winsvega:
I could create one using solidity test ipcsocket

chriseth:
Solidity's ipcsocket will not be available for a test for libethereum and we have quite some javascript tests already (they are just not run).

winsvega:
I mean I could test modifytimestamp feature from solidity hardcoded test using IPC socket.
where are the javascript tests?

PR #294, opened 2016-06-17
Exit on fatal error.

chriseth:


bobsummerwill:
Functionality looks fine.    Tests show as broken.    ?

chriseth:
Yes, I think we still have incorrect expectations as part of the tests. We can the generator on a machine with not enough ram (and those assertions not in place).
This should be merged and the test expectations updated after the merge taking care that these assertions are not triggered (by moving tests that take too much ram to the "we have enough ram" test suite).

chfast:
I would like the exception to be thrown higher and caught by any tool that uses libethereum.

chriseth:
@chfast the higher code is probably not exception-safe.

winsvega:
making testeth exit in this scenario would make running some tests impossible
If I run memoryStress tests on low memory machine It would just crash the whole execution
It would be better with std::error notification

chriseth:
@winsvega but what is the point in running the memory stress test on a low-memory machine?

winsvega:
such as my own. I am unable to fill such tests

chriseth:
@winsvega but then you should not do that on your machine, the tests would be incorrectly filled.
Please ask @jpitts for access to a suitable machine.

PR #295, opened 2016-06-20
add VM exception to the error log

winsvega:
move such tests under --all option

### DIFF
--- a/test/libethereum/StateTestsFiller/Homestead/stBoundsTestFiller.json
+++ b/test/libethereum/StateTestsFiller/Homestead/stBoundsTestFiller.json
@@ -320,50 +320,6 @@
         }
     },
 
-    "MLOAD_Bounds" : {
-        "env" : {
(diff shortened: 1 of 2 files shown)


libethereum/Executive.cpp
```diff
@@ -404,11 +404,13 @@ bool Executive::go(OnOpFunc const& _onOp)
 		catch (Exception const& _e)
 		{
 			// TODO: AUDIT: check that this can never reasonably happen. Consider what to do if it does.
+			std::cerr << "VM ERROR!" << endl;
```
chriseth: Please exit the process (I have an open PR for that) - we should not continue in this state.

chriseth:
Please merge as soon as tests (except for elcapitan) are green.

winsvega:
ok. so this PR is just changing the test creation logic. (not execution or run)
The test results under this PR actually show how latest develop tests are run on Jenkins

chriseth:
Argh, build error!

PR #299, opened 2016-06-28
remove old fields form vmtrace

yann300:


chriseth:
Is the code to retrieve the missing fields already in place in remix?

yann300:
Yes and those fields are not in the geth vmtrace

PR #301, opened 2016-06-28
EVMJIT: Use EVM-C callback functions

chfast:
EVMJIT PR: https://github.com/ethereum/evmjit/pull/67


libevm/JitVM.cpp
```diff
@@ -14,11 +13,135 @@ namespace dev
 {
 namespace eth
 {
+namespace
+{
+
+static_assert(sizeof(Address) == sizeof(evm_hash160),
```
chriseth, line 19: I kind of remember that static_assert is not available on all our platforms. That might have changed if we move to a newer msvc version.

libevm/JitVM.cpp
```diff
@@ -14,11 +13,135 @@ namespace dev
...
+              "Address types size mismatch");
+static_assert(alignof(Address) == alignof(evm_hash160),
+              "Address types alignment mismatch");
+
+evm_hash160 toEvmC(Address _addr)
```
chriseth: Wouldn't it be better to pass by const ref here (also below)?

libevm/JitVM.cpp
```diff
@@ -14,11 +13,135 @@ namespace dev
...
+	// TODO: h256 should be aligned to 8 bytes.
+	evm_hash256 g;
+	std::memcpy(&g, &_h, sizeof(g));
+	return g;
+}
+
+evm_uint256 toEvmC(u256 _n)
+{
+	evm_uint256 m;
+	m.words[0] = static_cast<uint64_t>(_n);
```
chriseth, line 52: Better to use `_n.convertTo<uint64_t>()`?

libevm/JitVM.cpp
```diff
@@ -14,11 +13,135 @@ namespace dev
...
+	u <<= 64;
+	u |= _n.words[2];
+	u <<= 64;
+	u |= _n.words[1];
+	u <<= 64;
+	u |= _n.words[0];
+	return u;
+}
+
+evm_variant evm_query(evm_env* _opaqueEnv, evm_query_key _key,
```
chriseth: If you don't mind:

```
evm_variant evm_query(
    evm_env* _opaqueEnv,
    evm_query_key _key,
    evm_variant _arg
) noexcept
```

libevm/JitVM.cpp
```diff
@@ -14,11 +13,135 @@ namespace dev
...
+			break;
+		case EVM_TIMESTAMP:
+			// TODO: Handle overflow / exception
+			v.int64 = static_cast<int64_t>(env.envInfo().timestamp());
+			break;
+		case EVM_CODE_BY_ADDRESS:
+		{
+			auto addr = fromEvmC(_arg.address);
+			auto &code = env.codeAt(addr);
+			v.bytes.bytes = reinterpret_cast<char const*>(code.data());
```
chriseth: Is the lifetime of code long enough? I think it might be invalidated as soon as we do a recursive create, because that might cause a re-allocation of the m_cache map.

libevm/JitVM.cpp
chfast, line 19: I pretty sure static_assert is supported by VS 2013. However, I've seen some bugs there. Nevertheless, I hope we will be able to test it in the CIs. In case of trouble, we can disable that on Windows.

libevm/JitVM.cpp
chfast: I really hope the compiler will inline that, then it does not matter. But good point. I planned to check it somehow.

libevm/JitVM.cpp
chfast, line 52: Is it something different? If not I would prefer to keep generic expressions. How about exceptions here?

libevm/JitVM.cpp
chfast: True, looks scary. It used to work that way before https://github.com/ethereum/libethereum/blob/develop/libevm/JitEnv.cpp#L106-L112.

I think evmjit requires that to be valid only for a moment because the opcode is EXTCODECOPY and as the name suggests it copies the memory.

But this is good point and this case has to be at least be documented in the EVM-C.

chfast:
Thanks very much for reviewing that. It is still work in progress. But that helped me identified some issues in the EVM-C API. I will also need to introduces more changes to the EVM-C.

I planned to merge this when all the callbacks are replaced with the ones specified in the EVM-C. I hope it will no be very big then. But we can discuss this approach.

libevm/JitVM.cpp
chriseth, line 52: Ok, let's keep it like that.

libevm/JitVM.cpp
chfast: I changed that but I think we should give up formatting that cannot be done by any tool. I tried to force clang-format to style like that, no success so far :)

gcolvin:
@chfast, I looked over the changes, especially the additions to JitVM.cpp.  They look good to me.

chfast:
@chriseth would you like to take a look again (not pushing)?

When is the best time to merge it?

bobsummerwill:
We can merge whenever we are comfortable with the changes.

I don't think we need to hold-off until after 1.3.1, for example, @chfast.

It's trivial for me to keep my cpp-ethereum in sync with webthree-umbrella changes as they come in.    Don't feel that we need to stop working :-)

chfast:
Not merging that is not blocking me. However, we don't any CI with evmjit enabled.

bobsummerwill:
Right - because of the build times for LLVM.

chriseth:
If this 'works' (choose your definition), it can be merged.

PR #302, opened 2016-06-28
disable failing block tests

winsvega:


chriseth:
Please use line-level comments to disable something, so it is clearly visible that it is disabled.

chriseth:
Travis was stuck in the "output" test.

winsvega:
no that is bad. why Travis hang when a block imported into chain?

PR #306, opened 2016-06-29
Sleep if block is sealed but not yet ready for import.

chriseth:
This hopefully fixes the dreaded "invalid operation on sealed block" error. We occasionally still get mining timeouts from solidity tests, but they at least do not void the whole chain.

### DIFF
--- a/libethereum/Client.cpp
+++ b/libethereum/Client.cpp
@@ -631,7 +631,8 @@ void Client::rejigSealing()
 			if (wouldSeal())
 			{
 				sealEngine()->onSealGenerated([=](bytes const& header){
-					this->submitSealed(header);
+					if (!this->submitSealed(header))
+						clog(ClientNote) << "Submitting block failed...";
 				});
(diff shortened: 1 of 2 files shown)


libethereum/Client.cpp
```diff
@@ -631,7 +631,8 @@ void Client::rejigSealing()
 			if (wouldSeal())
 			{
 				sealEngine()->onSealGenerated([=](bytes const& header){
-					this->submitSealed(header);
+					if (!this->submitSealed(header))
```
winsvega, line 634: wont this throw an exception in Block.cpp that would lead to crash ?
https://github.com/ethereum/libethereum/blob/develop/libethereum/Block.cpp#L302

chriseth, line 634: I don't see how execution could get to that line from here on.

chriseth:
Will merge this PR for now, we can still make modifications if necessary.

PR #308, opened 2016-06-29
Do not include uncles twice.

chriseth:


chfast:
The code looks good. But is hard to follow the logic.

winsvega:
uncle tests seems to be fine

chfast:
Is it possible to add a test that shows the problem this change fixes? This is a good practice to add regression tests with bugfixes.

winsvega:
so chriseth the problem was that same uncle was added to some block twice? 
actually we have that test for uncle being included twice 
https://github.com/ethereum/libethereum/blob/develop/test/libethereum/BlockchainTestsFiller/Homestead/bcUncleTestFiller.json#L1186

and we have test for uncle already included as a block in chain 
https://github.com/ethereum/libethereum/blob/develop/test/libethereum/BlockchainTestsFiller/Homestead/bcUncleTestFiller.json#L422

test just check that such block would throw an exception. 
this fix should fix the reason such situation occurred in node

chriseth:
The tests you mentioned, dimitry, were fine because the mechanism to detect multiple uncles is correct. The problem was that the same uncle was included twice as part of the block generation. The root cause for that was, I think, how we mess with the blockchain in solidity tests. I think under normal circumstances, this can never happen. I also added a second commit that prevents such database corruption.

@chfast adding a test is probably quite hard and, as I said, this does not happen for normal chains, but it will happen for our solidity tests, so we are fine there, I think.

chfast:
:ok:

PR #309, opened 2016-06-30
Travis CI: whitelist only master and develop branches

chfast:


bobsummerwill:
Where is the other build coming from, @chfast?
You did mention this elsewhere, but I have forgotten the specifics.

chfast:
Travis has 2 build triggers: push and pr. If you pushed a branch "example-feature" to libethereum repo it would trigger both.

chfast:
See https://github.com/ethereum/libethereum/pull/301

bobsummerwill:
Thanks for the info.   Yeah - that makes total sense now.

PR #311, opened 2016-07-04
Environment variable support in ethminer

kost:
Initial support of environment variables in ethminer.
For the start, you can specify URL and type of mining
via environment variables.

### DIFF
--- a/ethminer/MinerAux.h
+++ b/ethminer/MinerAux.h
@@ -91,6 +91,12 @@ inline std::string credits()
 	return out.str();
 }
 
+std::string getEnvVar(std::string const& key)
+{
+    char const* val = getenv(key.c_str());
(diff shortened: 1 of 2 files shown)


bobsummerwill:
Hey @kost,

It would probably be better for you to submit this PR to @Genoil against http://github.com/Genoil/cpp-ethereum.

The official ethminer hasn't really been touched for months, but we're planning to upstream from Genoil https://github.com/ethereum/webthree-umbrella/issues/315.    Given that plan, I don't really want to make any changes to our ethminer in the meantime.

Best wishes!

bobsummerwill:
Closing, because the repos have all been moved around.

If you have not already done so, @kost, I would urge you to make this PR against the @Genoil miner, and for it to make its way back here through the future upstreaming of that codebase.   Thanks!

PR #314, opened 2016-07-11
CMake: make EVMJIT available only on x86_64

chfast:
The changes in EVMJIT are in https://github.com/ethereum/evmjit/pull/68 but they can be merged independently.

Supports https://github.com/ethereum/webthree-umbrella/issues/626.


bobsummerwill:
LGTM - please merge both PRs whenever you like, and thank you :-)

PR #315, opened 2016-07-12
Deleted unused libnatspec library

bobsummerwill:
Moved libtestutils into the tests\ folder.


libnatspec/natspecjs/package.json
```diff
@@ -1,24 +0,0 @@
-{
```
chriseth, line 1: Wait a second, I was not aware that this is an npm package. Perhaps we should check again whether this is in use somewhere.
Perhaps we should extract the javascript part somewhere else and just delete the cpp part.

bobsummerwill, line 1: I've created a new issue for that future TODO:
https://github.com/ethereum/webthree-umbrella/issues/648

gcolvin:
It's trivial for somebody to grab deleted code back out of the PR for now, on the slim chance npm is needed. Can consider further in #648.

PR #316, opened 2016-07-13
fix current pc

yann300:


chfast:
It's hard to tell it is right. Can you explain it?

Shouldn't the VM produce incorrect results if the PC is wrong? If yes, you should provide a regression test that proves the bug in the VM.

bobsummerwill:
I agree.   We need tests around EVM changes.   CC @winsvega, @chriseth.

@gcolvin's changes were big and scary but they were refactorings, not bug-fixes or changes.    They might have broken something, because we won't really know how much coverage we have, but at least they didn't break any tests.

What is the incorrect behaviour you are seeing, Yann?   How did this manifest for you?

yann300:
i've rebased not so long ago (like 5 days). Before, the associated current PC for a call (or create) opcode given by the vmtrace was matching with the executing code. I mean that the pc was pointing to the right location in the executing code. Which is not the case now. the pc is incremented one time more now.

chriseth:
@bobsummerwill we certainly have tests for these. I think we have 100% code coverage of the VM - which of course does not mean all cases are covered, just that all lines are touched, but christoph and dimitry did a very good job before frontier.

bobsummerwill:
I just don't understand why some existing tests aren't failing prior to the change ... ?

As @chfast says, it should be possible for us to have a regression test which fails, and then shows as fixed after this change is made, should it not?

chfast:
@bobsummerwill :+1: 
This should be a practice -- if a bug in critical component is spotted, you should first prepare a test that shows the bug, then apply a fix. Some say the test should be committed separately, I wouldn't be so pedantic. 

When the bug is not covered by any test so far, and no test failures after the fix is applied is simply not enough. Regression tests is a must. That's my opinion.

bobsummerwill:
Yes.   Post repo-reorg, I hope to actually be spending time in the C++ code fixing bugs, implementing features (LES, fast-sync, etc) and so on, and given that we have largely inherited the code, with the original authors working on other things, test completeness is pretty essential.

The kind of refactoring work which Greg did is very, very scary, and I am glad it was him, not me.

Any changes in the consensus code really do need tests, I think.

So, is the conclusion that we are missing unit-testing for the specific flaw which @yann300 is addressing here?    What badness is manifesting with the current code, Yann?   Or is just some internal inconsistency which you only see by virtual of new code in remix?

yann300:
I mean that the pc (of a CALL) that is given by the trace does refer to a CALL op code in the execution code (it refer to the next opcode).
afak the pc is only used to locate the op code in the execution code. So this is not involved and cannot cause bugs in the execution.
I will write a test that reveal this bug. @winsvega could you help me on that?

chriseth:
Please merge if VM tests pass.

PR #321, opened 2016-07-18
Populate lookups in KeyManager in masterless mode.

chriseth:


chfast:
Code looks good, but for me is hard to follow logic for this part. So if you are sure it is good fix and all tests pass -- go ahead and merge.

bobsummerwill:
Yes - I have much the same experience as chfast.    My review doesn't add much value. 

Merge if confident, @chriseth!

PR #322, opened 2016-07-18
Fix regression introduced by commit 80b97195.

lvella:
Commit 80b97195 enabled mining on Mesa's OpenCL implementation, while
breaking OpenCL mining for everyone else. This commit includes a
feature check around directive to enable exclusive Mesa extension.


lvella:
Please test with other OpenCL implementations.

bobsummerwill:
Many thanks!   Merged.    We'll tweak as required from here.

PR #323, opened 2016-07-22
update tests gasLimit, add randomTests as a fillers

winsvega:
update now wrong gasLimit in the old tests
add randomTests as a fillers


chriseth:
Please explain why this 28k additional lines are needed. How does that increase test runtime and coverage? Does it add some cases we did not test yet or is it just random tests?

winsvega:
the idea is to move as much of the test sources to libethereum testeth as possible. 
because tests are recreated by testeth and thus it might be not safe to keep the source in test repository.

at this time I had to make StateTests/RandomTests  cases (~680 tests) into a fillers because all of them has outdated and incorrect blockGasLimit. so now all the random tests would be refilled through the testeth

PR #325, opened 2016-07-25
fix keymanager addresses list

winsvega:
connects to https://github.com/ethereum/webthree-umbrella/issues/676

### DIFF
--- a/libethcore/KeyManager.cpp
+++ b/libethcore/KeyManager.cpp
@@ -422,13 +422,15 @@ void KeyManager::write(SecureFixedHash<16> const& _key, string const& _keysFile)
 {
 	RLPStream s(4);
 	s << 1; // version
-	s.appendList(accounts().size());
-	for (auto const& address: accounts())
+
+	s.appendList(m_keyInfo.size());
+	for (auto const& info: m_keyInfo)
 	{
-		h128 id = uuid(address);
-		auto const& ki = m_keyInfo.at(address);
-		s.appendList(5) << address << id << ki.passHash << ki.accountName << ki.passwordHint;
+		h128 id = uuid(info.first);
+		auto const& ki = info.second;
+		s.appendList(5) << info.first << id << ki.passHash << ki.accountName << ki.passwordHint;
 	}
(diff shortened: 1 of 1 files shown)


libethcore/KeyManager.cpp
```diff
@@ -422,12 +422,35 @@ void KeyManager::write(SecureFixedHash<16> const& _key, string const& _keysFile)
...
-	for (auto const& address: accounts())
+
+	int validAccountsSize = 0;
+	Addresses allAccounts = accounts();
+	for (auto const& address: allAccounts)
+	{
+		try
+		{
+			m_keyInfo.at(address);
+			validAccountsSize++;
```
winsvega: kind of awkward solution

libethcore/KeyManager.cpp
```diff
@@ -422,12 +422,20 @@ void KeyManager::write(SecureFixedHash<16> const& _key, string const& _keysFile)
...
 	s << 1; // version
-	s.appendList(accounts().size());
+
+	s.appendList(m_keyInfo.size());
 	for (auto const& address: accounts())
 	{
-		h128 id = uuid(address);
-		auto const& ki = m_keyInfo.at(address);
-		s.appendList(5) << address << id << ki.passHash << ki.accountName << ki.passwordHint;
+		try
```
chriseth: Don't use exceptions in C++ if you can avoid them, they are really expensive.
So please rather use something like: `if (!m_keyinfo.count(address)) continue;`

libethcore/KeyManager.cpp
```diff
@@ -422,12 +422,20 @@ void KeyManager::write(SecureFixedHash<16> const& _key, string const& _keysFile)
 {
 	RLPStream s(4);
 	s << 1; // version
-	s.appendList(accounts().size());
+
+	s.appendList(m_keyInfo.size());
 	for (auto const& address: accounts())
```
chriseth: I think it would be better to just iterate over `m_keyInfo` - this would also better match the previous line.

PR #326, opened 2016-07-26
Improve Travis

chfast:


bobsummerwill:
I'm guessing this is still WIP, eh?

bobsummerwill:
Are you doing anything with this, @chfast?

.travis.yml
```diff
@@ -1,5 +1,8 @@
 language: cpp
-cache: ccache
+cache:
+    - ccache
+    - directories:
+        - ~/.ethash
```
chfast, line 5: Caching DAGs seems good idea.

chfast:
This can be closed. I tried to workaround broken llvm package of Ubuntu.

The only valuable part is Travis config to cache ethash DAG.

bobsummerwill:
OK - why don't you make a new PR for the DAG caching against cpp-ethereum and then we're done with this?  Thanks!

PR #330, opened 2016-07-28
fix calculating incorrect gasLimit from genesis

winsvega:
fixing mining transactions on a testchain

### DIFF
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -1430,7 +1430,11 @@ Block BlockChain::genesisBlock(OverlayDB const& _db) const
 {
 	h256 r = BlockHeader(m_params.genesisBlock()).stateRoot();
 	if (_db.exists(r))
-		return Block(*this, _db, r);
+	{
+		Block gen(*this, _db, r);
+		gen.m_currentBlock = BlockHeader(m_params.genesisBlock());
+		return gen;
+	}
 	Block ret(*this, _db, BaseState::Empty);
 	ret.noteChain(*this);
 	dev::eth::commit(m_params.genesisState, ret.mutableState().m_state);	// bit horrible. maybe consider a better way of constructing it?
--- a/libethereum/BlockChain.cpp
+++ b/libethereum/BlockChain.cpp
@@ -1429,22 +1429,19 @@ bytes BlockChain::headerData(h256 const& _hash) const
 Block BlockChain::genesisBlock(OverlayDB const& _db) const
 {
 	h256 r = BlockHeader(m_params.genesisBlock()).stateRoot();
-	if (_db.exists(r))
-	{
-		Block gen(*this, _db, r);
-		gen.m_currentBlock = BlockHeader(m_params.genesisBlock());
-		return gen;
-	}
 	Block ret(*this, _db, BaseState::Empty);
-	ret.noteChain(*this);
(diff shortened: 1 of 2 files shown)


libethereum/BlockChain.cpp
```diff
@@ -1430,7 +1430,11 @@ Block BlockChain::genesisBlock(OverlayDB const& _db) const
 {
 	h256 r = BlockHeader(m_params.genesisBlock()).stateRoot();
 	if (_db.exists(r))
-		return Block(*this, _db, r);
+	{
+		Block gen(*this, _db, r);
```
chfast: Please name the variable `genesis`. `gen` suggests something generated.

libethereum/BlockChain.cpp
```diff
@@ -1430,7 +1430,11 @@ Block BlockChain::genesisBlock(OverlayDB const& _db) const
...
+		gen.m_currentBlock = BlockHeader(m_params.genesisBlock());
```
chfast: This expression is executed a couple lines above. Please do not evaluate it again.

chfast:
You provided wrong issue reference.

winsvega:
why?

chfast:
> fixing mining transactions on a testchain
> connects to #330

The link points to this PR :)

chriseth:
From the structure, it seems that this should be made differently. The check at the beginning is a shortcut that is taken if the state already exists. So I think something like the following should also work and might be better:

```
    h256 r = BlockHeader(m_params.genesisBlock()).stateRoot();
    Block ret(*this, _db, BaseState::Empty);
    if (!_db.exists(r)) {
        ret.noteChain(*this);
        dev::eth::commit(m_params.genesisState, ret.mutableState().m_state);    // bit horrible. maybe consider a better way of constructing it?
        ret.mutableState().db().commit();                                       // have to use this db() since it's the one that has been altered with the above commit.
        if (ret.mutableState().rootHash() != r)
        {
            ...
        }
    }
    ret.m_previousBlock = BlockHeader(m_params.genesisBlock());
    ret.resetCurrent();
    return ret;
```

winsvega:
chfast it seems that waffle dispays this PR as an issue so there is no issue I could connect this pr to

christian way seems to work too but differently 
/home/wins/Ethereum/umbrella/libethereum/test/libethereum/BlockChain.cpp(167): fatal error in "sync": critical check importIntoQueue == ImportResult::Success failed

that means hat some block was previously imported fine. now i's not not. 
test code:

```
TestBlockChain bc2(TestBlockChain::defaultGenesisBlock());
TestBlock block;
block.mine(bc2);
bc2.addBlock(block);

TestBlock block2;
block2.mine(bc2);

BlockQueue uncleBlockQueue;
uncleBlockQueue.setChain(bc2.interface());
ImportResult importIntoQueue = uncleBlockQueue.import(&block2.bytes(), false);
BOOST_REQUIRE(importIntoQueue == ImportResult::Success);
```

winsvega:
cant import block2 that mined on bc2 after block1  into blockqueue for that blockchain

hm and it seems that mining blocks with uncles now hang for unlimited time. 
not sure if it's related

---

no it is not related. some rare mining bug it seems. 
as for sync test it is unreliable. sometimes it fails sometimes it's not. strange. 
then I guess this commit is ok. all tests work

chriseth:
Please address pawel's comments.

winsvega:
I removed the reference

PR #331, opened 2016-07-28
consolidate exception throws

gcolvin:
Move throw of each kind of exception into its own non-inlined function so that all the boost exception code doesn't get inlined in interpreter cases.  @chfast

### DIFF
--- a/libevm/VM.cpp
+++ b/libevm/VM.cpp
@@ -37,30 +37,21 @@ inline u256 fromAddress(Address _a)
 	return (u160)_a;
 }
 
-template<class T> static uint64_t toUint64(T v)
-{
-	// check for overflow
-	if (v > 0x7FFFFFFFFFFFFFFF)
-		throwVMException(OutOfGas());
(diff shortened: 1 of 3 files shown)


gcolvin:
@bobsummerwill Ubuntu passes, Mac fails for lack of time.  Do we merge it anyway?

bobsummerwill:
Yes, no gating right now.

Have you done a local mac build, in case there are some silly build errors?

gcolvin:
No, my current Mac is inadequate.

gcolvin:
No way to give it more time?

bobsummerwill:
No way to give it more time.

Because yours is Mavericks only?    If so, good news when you're on cpp-ethereum in a few days, because I added support back in for Mavericks.

Just merge away.   We'll fix it up if anything breaks.

gcolvin:
Will merge.  I'm on Lion -- old machine is too wimpy for Yosemite.  I just use it as a smart terminal for my Ubuntu box.

PR #333, opened 2016-08-01
split sync test and limit mining time in the tests

winsvega:
mining limit for tests

### DIFF
--- a/test/boostTest.cpp
+++ b/test/boostTest.cpp
@@ -84,16 +84,6 @@ int main( int argc, char* argv[] )
 	//Initialize options
 	dev::test::Options const& opt = dev::test::Options::get(argc, argv);
 
-	char* path = (argv[0]);
-	argc = 3;
-	argv = new char*[3];
-	argv[0] = new char[strlen(path)+1];
-	std::strcpy(argv[0], path);
-	argv[1] = new char[3];
-	std::strcpy(argv[1], "-t");
-	argv[2] = new char[strlen("BlockChainSuite")+1];
-	std::strcpy(argv[2], "BlockChainSuite");
-
 	if (opt.createRandomTest)
 	{
 		//disable initial output
--- a/test/boostTest.cpp
+++ b/test/boostTest.cpp
@@ -83,6 +83,17 @@ int main( int argc, char* argv[] )
 {
 	//Initialize options
 	dev::test::Options const& opt = dev::test::Options::get(argc, argv);
+
+	char* path = (argv[0]);
+	argc = 3;
+	argv = new char*[3];
+	argv[0] = new char[strlen(path)+1];
+	std::strcpy(argv[0], path);
+	argv[1] = new char[3];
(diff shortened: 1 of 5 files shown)


test/boostTest.cpp
```diff
@@ -83,6 +83,17 @@ int main( int argc, char* argv[] )
 {
 	//Initialize options
 	dev::test::Options const& opt = dev::test::Options::get(argc, argv);
+
+	char* path = (argv[0]);
+	argc = 3;
```
chriseth: We should not just ignore the options that are passed here. Also I think it is really bad style to modify `argv`.

test/libethereum/BlockChain.cpp
```diff
@@ -60,17 +61,9 @@ BOOST_FIXTURE_TEST_SUITE(BlockChainSuite, TestOutputHelper)
 BOOST_AUTO_TEST_CASE(opendb)
 {
 	TestBlock genesis = TestBlockChain::defaultGenesisBlock();
-
 	TransientDirectory tempDirBlockchain;
 	ChainParams p(genesisInfo(Network::Test), genesis.bytes(), genesis.accountMap());
 	BlockChain bc(p, tempDirBlockchain.path(), WithExisting::Kill);
-
-	Block block = bc.genesisBlock(genesis.state().db());
```
chriseth, line 68: Does this still test enough if you remove these lines?

test/boostTest.cpp
winsvega: the only way I found it work without issues. It is just to test on travis only this test suite. 
will not be included to the final PR

test/libethereum/BlockChain.cpp
```diff
@@ -108,106 +101,210 @@ BOOST_AUTO_TEST_CASE(opendb)
...
+		{
+			BOOST_ERROR("Exception thrown when trying to mine or import a block!");
+		}
+	};
+
+	testMiningFunc(threadFunc);
 }
+
+// Temporary disable of "sync" test which is repeatedly failing in TravisCI for Ubuntu Trusty.
+#if !defined(ETH_AFTER_REPOSITORY_MERGE)
```
chriseth: Please remove this.

chriseth:
If this works, it might be a good workaround!

test/libethereum/BlockChain.cpp
winsvega, line 68: Yes. Just open same DB twice but without mining.

chriseth:
Still failing. I would like to see at least one successful run before we can merge it. Re-triggered travis.

winsvega:
bob has changed travis wait option to 40 minutes. it should work now

winsvega:
ok so I think it could be merged without last 2 commits (merge and remove)

bobsummerwill:
Make it so, @winsvega !

winsvega:
actually I already thought so 9 days ago )

PR #336, opened 2016-08-04
Rename asOf -> block

yann300:


yann300:
moved https://github.com/ethereum/cpp-ethereum/pull/3038

